PR #13, opened 2013-04-18
Send Emails when a calendar event is shared.

visitha:
I created a new feature to ownCloud Calendar app. That is to send emails for all shared people with few event details when a calendar event is shared.
Along with this pull request I have added all of my changes in the calendar repository. Please review these codes and give your suggestions to improve this feature.


lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+           foreach ($eventShareesNames as $name) {
+               $result = OC_Calendar_Calendar::getUsersEmails($name);
+               $emails[] = $result;
+           }
+           $useremail = OC_Calendar_Calendar::getUsersEmails($user);
+           foreach ($emails as $email) {
+               $subject = "Calendar Event Shared..!!";
+               $headers = "MIME-Version: 1.0\r\n";
+
+               $headers .= "Content-Type: text/html; charset=ISO-8859-1\r\n";
```
BernhardPosselt: I think you want utf-8 charset here :)

templates/part.share.php
```diff
@@ -53,6 +57,10 @@
 	print_unescaped('<div id="sharedWithNobody">' . OC_Util::sanitizeHTML($l->t('Nobody')) . '</div>');
 } ?>
 <br />
+<input type="button" style="float:right;" class="submit" value="<?php echo $l->t("Send Email"); ?>" onclick="Calendar.Util.sendmail('<?php echo $eventid;?>','<?php echo $location;?>','<?php echo $description;?>','<?php echo $dtstart;?>','<?php echo $dtend;?>');">
```
BernhardPosselt, line 60: please use p() instead of echo because heres an XSS vulnerability ;)

lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+
+               $headers .= "From:" . $useremail;
+
+               $message = '<html><body>';
```
BernhardPosselt: I also used to build emails like this some time ago. I think its better though if you put the html code into a template and use that to generate the body. Its easier to read and to find XSS vulnerabilities.

lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+   //$message .= '<img src="http://css-tricks.com/examples/WebsiteChangeRequestForm/images/wcrf-header.png" alt="Website Change Request" />';
```
BernhardPosselt: Why this image?

lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+        * @param $eventid The event id
+        * @param $location The location
+        * @param $description The description
+        * @param $dtstart The start date
+        * @param $dtend The end date
+        *
+        */
+       public static function sendEmails($eventid, $location, $description, $dtstart, $dtend) {
+
+           $user = OC_User::getUser();
```
georgehrke: please make use of our public api
OC_User::getUser(); -> \OCP\User::getUser();

lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+                   $eventShareesNames[] = $sharee['share_with'];
+               }
+           }
+           foreach ($eventShareesNames as $name) {
+               $result = OC_Calendar_Calendar::getUsersEmails($name);
+               $emails[] = $result;
+           }
+           $useremail = OC_Calendar_Calendar::getUsersEmails($user);
+           foreach ($emails as $email) {
+               $subject = "Calendar Event Shared..!!";
```
georgehrke: I guess "Calendar Event shared" is sufficient. No need for two periods and two exclamation points ;)

lib/app.php
```diff
@@ -502,4 +502,56 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+
+               $message = '<html><body>';
+   //$message .= '<img src="http://css-tricks.com/examples/WebsiteChangeRequestForm/images/wcrf-header.png" alt="Website Change Request" />';
+               $message .= '<table style="border:1px solid black;" cellpadding="10">';
+               $message .= "<tr style='background: #eee;'><td colspan='2'><strong>" . $user . "</strong><strong> has shared with you an event</strong></td></tr>";
+               $message .= "<tr><td><strong>Location:</strong> </td><td>" . $location . "</td></tr>";
+               $message .= "<tr><td><strong>Description:</strong> </td><td>" . $description . "</td></tr>";
+               $message .= "</table>";
+               $message .= "</body></html>";
+
```
georgehrke, line 551: please use single quotes instead of double qoutes
(see [coding guidelines](http://owncloud.org/contribute/) )

visitha:
@georgehrke @Raydiation Thank you very much both of you for pointing out those changes. Since this is my first pull request in ownCloud, there may be lot of mistakes. Now I am working on to correct those things. 
Can any one tell me about the format of HTML email template? Then we can create very attractive email with lot of details.
Thanks again..!

lib/app.php
visitha: If we want to add some images to our email, then we can use it. Therefore I kept it as commented since it may be needed in the future. BTW I will remove it for now.

BernhardPosselt: Ah, i see its a dead comment :) Sure remove it ;)

js/calendar.js
```diff
@@ -8,6 +8,25 @@
...
+			OC.filePath('calendar','ajax/event','sendmail.php'),
+			{
+				eventId:eventId,
+                                location:location,
+                                description:description,
+                                dtstart:dtstart,
+                                dtend:dtend
+			},
+			function(result){
+				if(result.status!='success'){
```
Niduroki, line 22: Try using `!==` please.

lib/app.php
```diff
@@ -502,4 +502,54 @@ public static function generateEventOutput(array $event, $start, $end) {
...
+       public static function sendEmails($eventid, $location, $description, $dtstart, $dtend) {
+
+           $user =  \OCP\User::getUser();
+           $eventsharees = array();
+           $eventShareesNames = array();
+           $emails = array();
+           $sharedwithByEvent = OCP\Share::getItemShared('event', $eventid);
+           if (is_array($sharedwithByEvent)) {
+               foreach ($sharedwithByEvent as $share) {
+                   if ($share['share_type'] == OCP\Share::SHARE_TYPE_USER || $share['share_type'] == OCP\Share::SHARE_TYPE_GROUP) {
```
Niduroki, line 524: Try using `===` please.

Also change the filemode back to `644` please.

lib/calendar.php
```diff
@@ -382,4 +382,20 @@ public static function generateTextColor($calendarcolor) {
 		$computation = ((($red * 299) + ($green * 587) + ($blue * 114)) / 1000);
```
Niduroki, line 382: Change the filemode back to `644` please.

georgehrke:
moved to https://github.com/owncloud/calendar/pull/162

PR #15, opened 2013-04-19
Move calendar item to a shared calendar

patricks:
I had the problem described in https://github.com/owncloud/apps/issues/733

With this little fix, everything works fine.


georgehrke:
thx :)

PR #37, opened 2013-05-16
make the action buttons honour border in popup

simonbuehler:
hi,

previously the buttons were on a span and were misaligned from the left&right border, now they have the same distance from the popup border


georgehrke:
thx

PR #39, opened 2013-05-18
Use event id instead of calender id (issue #32)

mape2k:
Object::getowner returns the owner of an event, not of a calendar. So we have do use event id instead of calender id  - this was a fault in my first feature-adding patch and only affects access via CalDAV. This patch should fix the issue #32.


georgehrke:
thx

PR #40, opened 2013-05-18
Share also confidential events (fixes issue #5)

mape2k:
This patch allows to share confidential events (#5), only private or unknown class will not be shared.


georgehrke:
thx

PR #97, opened 2013-07-16
Change background color

dupondje:
Make the background color for 'today' a bit better visible.


MorrisJobke:
cc @georgehrke @bartv2: @dupondje has asked for state on this in IRC

PR #114, opened 2013-08-09
add: option to import single event share

ziirish:
Hi,

I've been following [this issue](https://github.com/owncloud/apps/issues/203) for several weeks.
And today I got this idea:

If I want to see single events share via caldav, I have to create an event using the shared informations.

So here it is, I added a new 'import' button in the 'showevent' view so that you can create your own event using the shared informations.

It is not the best we can do, but I hope it's a start.

### DIFF
--- a/ajax/event/import.form.php
+++ b/ajax/event/import.form.php
@@ -1,0 +1,282 @@
+<?php
+/**
+ * Copyright (c) 2011 Bart Visscher <bartv@thisnet.nl>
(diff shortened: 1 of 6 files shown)


jancborchardt:
Hey @ziirish, sorry for the late reply! It seems that the import option should be present in the settings and the newevent template rather than the showevent one. Right?

cc @georgehrke

jancborchardt:
I’m closing this because of no response to my question above.

We can of course reopen. :) Just answer the comment.

PR #153, opened 2013-09-24
Fix of #860: Code shown instead of HTML entity when opening timepicker

ariellp:


georgehrke:
thx for your contribution

:-)

j-ed:
This problem has not been fixed in OC 5.0.12 yet although the patch has been contributed several days before its publication date :-(

20131007_oc_calender_day_picker_html_code

PR #155, opened 2013-09-27
More robust event sharing - fixes 142

Gomez:
I got two databases where all shared events disappeared. I cant reproduce how this happened, but with this fix the broken events  get filted and a msg goes into the log. All working events are visible. 

owncloud/apps/calendar/ajax/events.php#33 fails because it gets a incomplete array.

The problem is, that events are oc_shared but not in oc_clndr_objects. 

Hope this helps, and we find a way to reproduce and the real problem.

This fix can go into master, too.

### DIFF
--- a/lib/share/event.php
+++ b/lib/share/event.php
@@ -30,16 +30,16 @@ class OC_Share_Backend_Event implements OCP\Share_Backend {
 	public function formatItems($items, $format, $parameters = null) {
 		$events = array();
 		if ($format == self::FORMAT_EVENT) {
-            foreach ($items as $item) {
-                $event = OC_Calendar_Object::find($item['item_source']);
-                if ($event == False) {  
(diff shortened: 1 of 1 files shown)


georgehrke:
thx a lot :)
I'll take a look and merge it as soon as possible

georgehrke:
master: 
20:38:07 georgehrke:~/devsrv/apps3/calendar > git lg
- 2efca8e - (HEAD, master) Fix indent, please (6 seconds ago) <Steffen Lindner>
- 6231c2b - Fix indent (16 seconds ago) <Steffen Lindner>
- 17cd9b0 - Make event sharing more robust - Helps Fix #142 (35 seconds ago) <Steffen Lindner>

Gomez:
This hasn't found is way into oC 5.0.13. 

There is the apps repo in owncloud/apps and here in calender. For releases the apps repo is used, does calender/stable5 and apps/stable5 needs a sync?? Or what's the preferred way?

PR #160, opened 2013-09-30
Port OC_VCategories to OCP\ITags and add CalDAV birthday calendar

tanghus:
@georgehrke sorry for having both things in the same branch. I forgot to separate them when coding


georgehrke:
:+1: thx

PR #163, opened 2013-10-03
Cast translated displayname to string in CalDAV bday calendar

tanghus:
I just noticed an error in my previous PR that made actual sync fail with:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre_DAV_Exception</s:exception>
  <s:message>Unknown property value type: object for property: {DAV:}displayname</s:message>
  <s:sabredav-version>1.7.6</s:sabredav-version>
</d:error>
```

@georgehrke


georgehrke:
Looks good, go ahead and merge it :+1:

Please excuse my brevity and typos.

PR #180, opened 2013-10-17
Suppress sessions

Cymaphore:
Ability to supress session handling in dav requests, see https://github.com/owncloud/core/issues/5383 ... I would recommend to add this to the 5.x-fork as well.

### DIFF
--- a/caldav.php
+++ b/caldav.php
@@ -1,4 +1,5 @@
 <?php
+define("DISABLE_SESSION", 1);
 if(!file_exists('../../lib/base.php')) {
 	die('Please update the path to /lib/base.php in caldav.php or make use of /remote.php/caldav/');
 }


caldav.php
```diff
@@ -1,4 +1,5 @@
 <?php
+define("DISABLE_SESSION", 1);
```
Niduroki, line 2: What if we don't want to disable sessions?

I'd say this should be a config.php setting, for those who need it.

jancborchardt:
@DeepDiver1975 @ringmaster your calls on this session handling solution please.

DeepDiver1975:
@jancborchardt we need to discus this in the core issue: https://github.com/owncloud/core/issues/5383

DeepDiver1975:
> @jancborchardt we need to discus this in the core issue: owncloud/core#5383

we need to walk a different path

PR #212, opened 2013-11-18
Fix calendar annoyances

### LINKED ISSUE #771: Weird scrolling behavior in calendar views
1st)
In the top middle of the window there are arrows to jump 1 week/month forward or backward, between the arrows is the current date-range. But these arrows and date-range is popping around every time I jump forward or backward, because a) there is an animated icon to show that the view is being refreshed for half a second on the right of the list-button and b) the size of the date-range is not fixed. My main-problem is, that the arrows are _not_ always on the same spot. So if I want to click the forward button repeatedly for scrolling several weeks, I sometimes miss the arrow because its position changed. There should be a fixed position for the arrows and a fixed width for the range-field.

2nd)
When using the scroll wheel on my mouse in the calendar, I want to just scroll the time-range of the current week, because not all of the 24 hours are fitting on the screen. But if I scroll to much, ownCloud scrolls to the next/previous week, which drives me crazy. Maybe we could have this behavior optional (user pref)

(ownClod 5.0 German/Google Chrome 25)

### LINKED ISSUE #1482: calendar: multiple layout issues
Version: 6.0 beta3

See the attached screenshot where I marked various layout issues.

1/ The elements are not properly centered and don't have the correct padding. They seem to small in comparison to other elements.

2/ The anchor should be centered.

3/ The elements are not properly aligned vertically.

4/ The elements are not properly aligned vertically (more visible with German locale)

jancborchardt:
A lot of fixes … cc @owncloud/designers @georgehrke 
- simplified the event detail view to focus on title and time
- fixed some layout issues cc @mbiebl, fix https://github.com/owncloud/apps/issues/1482
- remove superfluous tipsy popup which just repeated obvious info
- remove scrolling to switch moth which made the calendar unusable with lots of events, fix #24, fix https://github.com/owncloud/apps/issues/771
- make event detail view and settings view scrollable, fix #81
- color weekend, don’t bold weekday names, lighter border color
- display full share permissions in share tab
- disable half-baked list view, change for proper day view, fix #80 
- fix access class wording, fix #113 
- remove superfluous 'Active' checkbox in edit calendar view
- lots more details

Would be cool if someone of you can fix #185 (loading feedback not visible). That I couldn’t manage.

### DIFF
--- a/ajax/changeview.php
+++ b/ajax/changeview.php
@@ -12,7 +12,7 @@ $view = $_POST['v'];
 switch($view) {
 	case 'agendaWeek':
 	case 'month';
-	case 'list':
+	case 'agendaDay':
 		break;
 	default:
 		OCP\JSON::error(array('message'=>'unexpected parameter: ' . $view));
--- a/css/style.css
+++ b/css/style.css
@@ -134,3 +134,19 @@ button.category{margin:0 3px;}
 	font-size: 20px;
 	margin: 0;
 }
+
+.ui-dialog {
+	height: 94% !important;
+	top: 4% !important;
+	overflow-y: auto;
+	overflow-x: hidden;
+	padding: 0;
+}
+.ui-dialog .ui-dialog-content {
+	overflow: none;
+}
+#appsettings_popup {
+	height: 70%;
+	overflow-y: auto;
+	overflow-x: hidden;
+}
--- a/css/style.css
+++ b/css/style.css
@@ -122,3 +122,15 @@ button.category{margin:0 3px;}
 .link{
 	color:#148cf6;
 }
+
+
+
+
+
+#event-title {
+	-moz-box-sizing: border-box;
+	box-sizing: border-box;
+	width: 100%;
+	font-size: 20px;
+	margin: 0;
+}
--- a/css/style.css
+++ b/css/style.css
@@ -135,19 +135,72 @@ button.category{margin:0 3px;}
 	font-weight: bold;
 }
 
-/* bigger event title in detail view */
-#event-title {
+/* input fields take whole width */
+#event-title,
+#event-location,
+#category,
+#event-description {
 	-moz-box-sizing: border-box;
 	box-sizing: border-box;
 	width: 100%;
+	margin: 3px 0;
+}
+/* bigger event title in detail view */
+#event-title {
 	font-size: 20px;
+}
+/* group from and to date fields */
+#event-time {
+	display: block;
+	margin: 5px 0;
+}
+#event-time-from {
+	float: left;
+}
+#event-time-to {
+	float: right;
+}
+#from, #to {
+	width: 65px;
 	margin: 0;
 }
+#fromtime, #totime {
+	margin: 0;
+}
+#event-allday {
+	display: inline-block;
+	margin: 7px 0;
+}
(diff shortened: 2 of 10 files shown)


jancborchardt:
Cc @DeepDiver1975 @karlitschek @tanghus as well. You might like this :)

Also @simonruber @sabrina12 this fixes some of the calendar stuff we talked about!

MorrisJobke:
- for me the scrollbar is always present in the event view (popup window after click on an event) - even if I resize the window (Firefox, Chromium)
- the event list is gone (I think it's good to have a list of the upcoming event (day has appeared)

Otherwise: Great!

Possible things to do: remove shadow and decrease corner radius in settings menu

jancborchardt:
@kabum 
- scrollbar always present: Yeah, can you help me with that? Also do you know how to make the window non-resizable and non-movable? Some parameters to call the dialog with?
- yes, the agenda view went away because it looked ugly. No dragging possible neither.

Will look into the suggestions.

MorrisJobke:
> scrollbar always present: Yeah, can you help me with that? Also do you know how to make the window non-resizable and non-movable? Some parameters to call the dialog with?

~~Sorry. I'm not familiar with the dialog stuff.~~ Fixed

> yes, the agenda view went away because it looked ugly. No dragging possible neither.

Ugly is not an argument as you are a designer ;) Make it fancy. I don't think that it is good to remove such a
function. There you can have a overview of upcoming events without having them shortened because they have to fit in a tiny box.

**Edit:** I've read #80 and I'm fine with the day view as a replacement for the list.

jancborchardt:
The »remove shadow and decrease corner radius in settings menu« is fixed in core with https://github.com/owncloud/core/pull/5968 – please review.

jancborchardt:
Ok, fixed some other small popup issues. Please review again @owncloud/designers @georgehrke @mbiebl @surak @HolgerAusB

georgehrke:
What's about the list view?

jancborchardt:
As described above: The list view was changed to the day view. The list view, unlike the other views, looks very raw, does not support drag&drop and such. It’s custom, not from upstream fullcalendar, right? If we do a list view, we should work with upstream.

georgehrke:
> It’s custom, not from upstream fullcalendar, right?

Right. It was implemented by @bartv2

georgehrke:
@ everyone Are there any objections left?

MorrisJobke:
I'm okay with this :+1:

georgehrke:
:+1:

mbiebl:
@jancborchardt are those fixes in v6 beta5? I just tried that latest beta and all the issues I reported in owncloud/apps#1482 are still unfixed with that version.

mbiebl:
@jancborchardt nvm, apparently your branch was merged after beta5. So I applied the changes on top of beta5.
regarding the points I mentioned in owncloud/apps#1482
1/ is fixed now
2/ the dropdown is still broken, the arrow is now completely gone, and it now has has a scrollbar, i.e. the dialog doesn't properly resize.
3/ seems to be unfixed
4/ this dialog has been completely re-designed, so the issues mentioned here are no longer valid

MorrisJobke:
@mbiebl see https://github.com/owncloud/apps/issues/1482#issuecomment-29071056

PR #215, opened 2013-11-20
Update calendar.js

libasys:
Some small fix on edit an existing calendar. The old active variable checked the length of an checkbox ... so the return value was "2". Make this sense? Here's the fix!


georgehrke:
sounds reasonable :)
I'll give it a test later

georgehrke:
but why not set:

```
var active = $("#edit_active_"+calendarid).is(':checked');
```

?

libasys:
that's easy: the return value is true or false, but not 1 or 0 and in the DB there is an integer defined. And the values gets per ajax request and so you have a string(true/false).

PR #218, opened 2013-11-21
More calendar fixes

jancborchardt:
- clarify non-shared message, also add for shared via calendar
- fix share dropdown being hidden and hiding calendar which will be shared
- remove confusing note, instead only make it possible to edit calendar sharing settings in the calendar settings cc @te-online
- only append (group) indicator to groups, no need to put (user) after every user
- improve look of sharing tab
- remove 'can delete' permission, also resolve CSS 'delete' class conflict cc @DeepDiver1975 @georgehrke 

Please review @georgehrke @owncloud/designers – Georg, I’m not sure if I did the removal of »can delete« correctly.

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -577,7 +577,6 @@ Calendar={
 								+ '<span class="shareactions">'
 								+ '<label><input class="update" type="checkbox" checked="checked">'+t('core', 'can edit')+'</label>'
 								+ '<label><input class="share" type="checkbox" checked="checked">'+t('core', 'can share')+'</label>'
-								+ '<label><input class="delete" type="checkbox" checked="checked">'+t('core', 'can delete')+'</label>'
 								+ '<img class="svg action delete" title="Unshare"src="'+ OC.imagePath('core', 'actions/delete.svg') +'"></span></li>';
 							$('.sharedby.eventlist').append(newitem);
 							$('#sharedWithNobody').remove();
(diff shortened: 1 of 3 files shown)


MorrisJobke:
Looks great :+1:

te-online:
Cool :+1:

jancborchardt:
Thanks for the reviews!

I’d like to have a last look from @DeepDiver1975 or @georgehrke or someone who can verify that I removed the »can delete« correctly.

js/calendar.js
```diff
@@ -594,10 +596,9 @@ Calendar={
 						var permission = null;
 						if($(this).hasClass('update')) {
 							permission = OC.PERMISSION_UPDATE;
+							permission = OC.PERMISSION_DELETE;
```
DeepDiver1975, line 599: I guess it should be update and delete?

+= would help from my understanding

DeepDiver1975:
:-1: it's loosing the checked 'can edit' state

jancborchardt:
> it's loosing the checked 'can edit' state

Also happens with master, see https://github.com/owncloud/calendar/issues/217

DeepDiver1975:
> Also happens with master, see #217

in this case :+1:

js/calendar.js
jancborchardt, line 599: Oi, you merged, but what about this then? ;)

DeepDiver1975, line 599: this part has no influence at al to the behavior - this is the code piece which is broken and causes the check box not to be set properly

PR #226, opened 2013-11-22
Update new.form.php

libasys:
Added init values to prevend user input errors


georgehrke:
thx :)

ajax/event/new.form.php
```diff
@@ -79,8 +79,24 @@
 $tmpl->assign('endtime', $end->format('H:i'));
 $tmpl->assign('allday', $allday);
 $tmpl->assign('repeat', 'doesnotrepeat');
+
+//init hidden values date values for repeating
```
libasys, line 83: here comes the init for the translation

```
$l=OC_L10N::get('calendar');
```

PR #231, opened 2013-11-25
Fix for long timed events

libasys:
Fix for https://github.com/owncloud/calendar/issues/182
Also adding the request of objecttype, is feeling smarter. So only the vevents loaded (without vtodo) ;)


georgehrke:
thx :)

libasys:
via caldav the function all() is the preferd method

PR #252, opened 2013-12-10
remove cache setting, do not bother people with options like these

jancborchardt:
If at all, this cache should be automatically cleaned. This should not be a setting.

Please review @georgehrke @owncloud/designers

### DIFF
--- a/templates/settings.php
+++ b/templates/settings.php
@@ -95,15 +95,6 @@
 				</select>
 			</td>
 		</tr>
-		<tr class="advancedsettings">
-			<td>
-				<label for="" class="bold"><?php p($l->t('Cache'));?></label>
-				&nbsp;&nbsp;
-			</td>
-			<td>
-				<input id="cleancalendarcache" type="button" class="button" value="<?php p($l->t('Clear cache for repeating events'));?>">
-			</td>
-		</tr>
 	</table>
 </div>
 <h2 id="title_urls"><?php p($l->t('URLs')); ?></h2>


PVince81:
:+1: 

Never used that, no idea what it is...

georgehrke:
this is necessary for some users, we should keep it. it's in the settings, so users don't see it all the time

jancborchardt:
Can you explain »necessary for some users«? I have not seen that in any other calendar web app. Could we not have it be done automatically, much like expiring old versions?

libasys:
I am confirm with @georgehrke, this setting clears the repeating Table in the database and fill the repeating calendar events.

jancborchardt:
Yeah, I don’t get why this is a button for users to press. This should be done automatically if anything.

georgehrke:
If we would do this automatically for all users at the same time it would !@#$ the server

MorrisJobke:
@georgehrke ~~Then simulate the trigger of the button on "Save event".~~ Update: Sorry mixed that up. I thought it is in the event view, but it's in the settings view.

jancborchardt:
If doing it automatically for _all_ users at the _same_ time would mess up the server, do it automatically but _not_ for all users at the same time?

jancborchardt:
@georgehrke what about my above comment? Like clearing it automatically when you use the calendar?

georgehrke:
can we please show the button when debug mode is enabled

jancborchardt:
@georgehrke yeah. Just add that in this pull request – I don’t know how to hook that up.

georgehrke:
okay, will add asap

georgehrke:
master: https://github.com/owncloud/calendar/commit/f7e426c09a556e9fa4100740ca0fab81bcc379a0
stable7: https://github.com/owncloud/calendar/commit/eea79827d2b6c350d1ca07dd2093d5143a599b42

jancborchardt:
Thanks!

PR #253, opened 2013-12-11
Update search classes

abrown:
@tanghus, I think this should fix the problem you describe in https://github.com/owncloud/core/pull/4553#issuecomment-25425550. What I did is set up a search provider and search result as they would be set up using the updated search system; I feel that it makes it simpler to read the code and paves the way for richer search results (more properties, typed search results) while leaving the 'text' property in. Let me know what you think.

The only issue is that we can't test this until the above pull request is merged (or at least rebased); I tried and it kept on logging something like "js file not found: script:tags formfactor: serverroot:C:/Users/andrew/Code/core at C:\Users\andrew\Code\core\lib\template\resourcelocator.php#45".


libasys:
What's about search for a pattern, I tested this:
I would like have all termins of today so i added this in the search class of the calendar:
#dt search patterns = search for all events of today

```
function search($query){ ...
$today= date('Y-m-d',time());
$allowedCommands=array('#dt'=>1);   
```

Then on the foreach of the objects:

```
$searchAdvanced=false;
if(array_key_exists($query,$allowedCommands) && $allowedCommands[$query]){
    if($query=='#dt'){
      $search=$object['startdate']; 
      if(substr_count($search,$today)>0){
        $searchAdvanced=true;
      }
     }
 }

if(substr_count(strtolower($object['summary']), strtolower($query)) > 0 || $searchAdvanced==true) {
```

so you can add more search patterns options especially for the calendar or tasks

screen_search

abrown:
@libasys, I like the idea. I don't know what @bartv2, @tanghus, and @georgehrke think, but it should probably go in a different PR. I would think the best way to implement this would be to extract dates/times from the search query with PHP's strtotime() which supports all types of relative date formats like 'today' or 'last week' (http://www.php.net/manual/en/datetime.formats.relative.php). This way the user wouldn't have to learn/memorize special codes. What do you think?

libasys:
ok i tested this:
new function:

```
public static function validateDate($Str){
       $Stamp = strtotime( $Str );
       $Month = date( 'm', $Stamp );
       $Day   = date( 'd', $Stamp );
       $Year  = date( 'Y', $Stamp );

      return checkdate( $Month, $Day, $Year );
  }
```

in the search function:

```
function search($query) {...
$isDate=false;
  if(strlen($query)>=8 && self::validateDate($query)){
    $isDate=true;
  }
```

in the inner foreach:

```
$searchAdvanced=false;

if($isDate==true && strlen($query)>=8){
    $tempQuery=strtotime($query);
     $checkDate=date('Y-m-d',$tempQuery);
     if(substr_count($object['startdate'],$checkDate)>0){
          $searchAdvanced=true;
      }
}
if(substr_count(strtolower($object['summary']), strtolower($query)) > 0 || $searchAdvanced==true) { ...
```

now you can search for date, in each common format!

Edit: tested following: "yesterday,today,tomorrow" in the searchbox and it works perfectly!

abrown:
@libasys, I put the functionality you mentioned, though I implemented it a bit differently. Do you want to take a look at the code and maybe test to make sure it works as intended?

libasys:
Here is a link to my repo on github: https://github.com/libasys/oc513/tree/oc513/apps/calendar
Here you can find my implementation & modifications of the search including with a working showevent!

abrown:
Rebased...

georgehrke:
@andrewsbrown Does this fix #522?

georgehrke:
> including with a working showevent!

@libasys A pull request would be appreciated!

abrown:
@georgehrke, I'm not exactly sure because this PR was from so long ago, but from what I can tell #522 looks like a different issue

butonic:
Actually, I think it does. It seems owncloud/contacts#300 and owncloud/calendar#253 never got merged

georgehrke:
@andrewsbrown 
- [x] please use tabs instead of 4 spaces and
- [x] squash the commits

thx :)

georgehrke:
@PVince81 @butonic @icewind1991 You reviewed https://github.com/owncloud/core/pull/4553, could you help to review this pr please :) thx

georgehrke:
fixes #522

appinfo/app.php
```diff
@@ -10,7 +10,8 @@
 OC::$CLASSPATH['OC_Connector_Sabre_CalDAV_Calendar'] = 'calendar/lib/sabre/calendar.php';
 OC::$CLASSPATH['OC_Connector_Sabre_CalDAV_CalendarObject'] = 'calendar/lib/sabre/object.php';
 OC::$CLASSPATH['OC_Calendar_Repeat'] = 'calendar/lib/repeat.php';
-OC::$CLASSPATH['OC_Search_Provider_Calendar'] = 'calendar/lib/search.php';
+// OC::$CLASSPATH['OCA\Calendar\Search\Provider'] = 'calendar/lib/search/provider.php';
```
icewind1991: Remove these lines if they are no longer needed

butonic:
since all issues have been fixed :+1:

georgehrke:
:+1:

georgehrke:
@butonic this is safe to back port to stable7, right?

PR #259, opened 2013-12-16
OC 6.0.0a - Fixed graphic glitch in "edit event" form

tsumi:
Added width to #fromtime and #totime classes in order to avoid a grapic glitch on "edit event" form in calendar (see screenshot for details)

calendar-glitch

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -195,6 +195,7 @@ button.category{margin:0 3px;}
 }
 #fromtime, #totime {
 	margin: 0;
+	width: 60px;
 }
 #event-allday {
 	display: inline-block;
--- a/css/style.css
+++ b/css/style.css
@@ -200,6 +200,7 @@ button.category{margin:0 3px;}
 #event-allday {
 	display: inline-block;
 	margin: 7px 0;
+	width: 100%;
 }
 /* category input field leaves room for edit button */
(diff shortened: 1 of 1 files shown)


georgehrke:
cc @jancborchardt

jancborchardt:
@tsumi yes, thank you! Meant to fix that as well, thank you for being faster. :) You’re also welcome to dive into the other design issues and see if you can fix anything. :)

jancborchardt:
@tsumi oooh wait actually it doesn’t fully work. Try it in Chrome – there’s up and down buttons inside the field. We need 70px or 80px for them to fit too. Can you change that?

tsumi:
@jancborchardt set to 80px, i havn't chrome installed now, can you do a quick?

jancborchardt:
@tsumi one other thing: In other languages where the »All day« event checkbox text is short, it’s displayed between the times. To fix that, you can add `width: 100%; float: left;` to `#event-allday`.

tsumi:
Right ... with some "short" language like Japanese this happen ...
Only `width:100%` is needed, with `float:left` same problem happen with Advanced option button :D
owncloud

jancborchardt:
Nice, thank you @tsumi!

PR #265, opened 2013-12-19
Event title is no longer selected after being focused.

ben-denham:
This fix for https://github.com/owncloud/calendar/issues/224 will make sure that the event title is focused (but not selected) when the event dialog is opened.
Solution based on: http://stackoverflow.com/a/12892564

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -136,6 +136,9 @@ Calendar={
 			$('#sendemailbutton').click(function() {
 				Calendar.Util.sendmail($(this).attr('data-eventid'), $(this).attr('data-location'), $(this).attr('data-description'), $(this).attr('data-dtstart'), $(this).attr('data-dtend'));
 			})
+			// Focus the title, and reset the text value so that it isn't selected.
+			var val = $('#event-title').val();
+			$('#event-title').focus().val('').val(val);
 		},
 		newEvent:function(start, end, allday){
 			start = Math.round(start.getTime()/1000);


georgehrke:
cc @jancborchardt

georgehrke:
We actually disabled the automatic focus on purpose

ben-denham:
Oh ok, it was still automatically focusing and selecting for me in 6.0.0a (stable) and on the master branch (the #event-title field has autofocus="autofocus" in templates/part.eventform.php). But I see now that autofocus has been removed in the rework branch.

georgehrke:
> But I see now that autofocus has been removed in the rework branch.

All the template / js code you can find in the rework branch will be replaced anyhow.

jancborchardt:
@georgehrke Doesn’t matter if the code gets replaced. The current version is in use in ownCloud 6 which we just released. We have no clue if we can get the rework done until ownCloud 7 – we wanted to have it for ownCloud 6 already. ;) Also, even though the code might be replaced, we should build a bit on the design we have now – not everything will just be thrown away.

jancborchardt:
@ben-denham aaand again, your fix works perfectly! Thank you so much! Would really love if you work on the Calendar some more if you’re interested. :)

ben-denham:
You're very welcome, and I'd be happy to work on the calendar app some more when I have the time :)

I've actually been working on a fix for https://github.com/owncloud/calendar/issues/21 and https://github.com/owncloud/calendar/issues/131, that would allow users to hide shared calendars and personalize their display names and colours. I'll make another merge request for this soon.

I also have a few other smaller bug fixes that I will make some merge requests for as well.

jancborchardt:
@ben-denham great to hear! Looking forward to working more with you and seeing more contributions. :)

PR #266, opened 2013-12-19
Month table no longer jumps when the event dialog is opened

ben-denham:
This fix for https://github.com/owncloud/calendar/issues/222 hides the dialog_holder so that its height doesn't increase when a dialog is loaded into it, which was causing the month table to jump as described in the issue.


georgehrke:
cc @jancborchardt

jancborchardt:
@ben-denham awesome fixes man! Works great!

@georgehrke again here the question, can we backport or do we not play the stable6 game here?

PR #268, opened 2013-12-20
Fixed share-icon overlapping glitch

ben-denham:
Fix for https://github.com/owncloud/calendar/issues/119 that uses background images instead of img elements for share icons, so that icons are replaced (instead of overlapped) when OC.Share.updateIcons() is called.


georgehrke:
cc @jancborchardt

jancborchardt:
@ben-denham nice fix! :) Tested and works here.

@georgehrke we should probably also backport to stable6, if there’s a process for this in the calendar.

PR #277, opened 2013-12-31
Fix non ascii character display on calendar share dialog

vgezer:
Before, sanitizeHTML for displaying the name of the calendar was being
encoded. This was causing non-ascii characters to be shown strange.
This should fix this problem. Since this is only a display text,
sanitizing should not be a problem.


vgezer:
ping

georgehrke:
@LukasReschke p() already sanitizes the output, so this should be XSS safe, right?

LukasReschke:
@georgehrke Yup

georgehrke:
Thanks for your contribution :)

PR #288, opened 2014-01-06
Added calendar-user preferences.

### LINKED ISSUE #21: setup caldavtester
http://calendarserver.org/wiki/CalDAVTester

ben-denham:
These changes allow a user to hide/show and change the display name and colour of a calendar that is shared with them, without affecting how the calendar is displayed to other users.

It does this by storing calendar-user preferences for each calendar a user has access to (whether it is owned by them or shared with them or a group they belong to) in a new "clndr_calendar_user_preferences" database table.

Therefore, the "cldnr_calendars.active" column has been removed, because that setting is now stored per user in "clndr_calendar_user_preferences.active". However, the "cldnr_calendars.calendarcolor" and "cldnr_calendars.displayname" have been retained to serve as default preferences for the calendar.

When a user loses access to a calendar (e.g. it is unshared with them), their preferences are retained in the database. That way, if the user regains access to a calendar that used to be shared with them, their old preferences will still be applied to the calendar.

Because retaining these unused preferences could create a lot of obsolete rows in the database, a new button has also been added in the calendar settings for administrators to "Delete unused calendar preferences", which will delete any preferences where the user no longer has access to the calendar.

This fixes https://github.com/owncloud/calendar/issues/21 and https://github.com/owncloud/calendar/issues/131, and was mentioned in https://github.com/owncloud/calendar/pull/265#issuecomment-31618521.


georgehrke:
Thanks
a few notes:
- I think we shouldn't remove the active field in clndr_calendars. Everytime we changed the databasestructure of existing tables, there was a major fuckup (just search the core repo for database migration issues)
- "_calendar_user_preferences" is too long for oracledb, please use smth shorter

@DeepDiver1975 What do you think about removing this field?

georgehrke:
and "_calendar_user_preferences" should be a key-value store

ben-denham:
Thanks for the feedback @georgehrke!

I've shortened the table name to clndr_user_preferences instead of clndr_calendar_user_preferences.

I have also refactored clndr_user_preferences to be a key-value store.

Would you like me to put the active field back in clndr_calendars now, or should I wait until @DeepDiver1975 has replied?

Cheers!

georgehrke:
> Would you like me to put the active field back in clndr_calendars now, or should I wait until @DeepDiver1975 has replied?

Let's wait for @DeepDiver1975 's opinion

jancborchardt:
Are there any design changes, i.e. is my review necessary?

ben-denham:
@jancborchardt I wouldn't say there are any design changes, other than the addition of the "Delete unused calendar preferences" button in the calendar settings.

jancborchardt:
@ben-denham hm, strange button. Do we need that? What does it do? Would a user ever need to click it – and if so, what for? :)

georgehrke:
> > @ben-denham hm, strange button. Do we need that? What does it do? Would a user ever need to click it – and if so, what for? :)
> 
> Because retaining these unused preferences could create a lot of obsolete rows in the database, a new button has also been added in the calendar settings for administrators to "Delete unused calendar preferences", which will delete any preferences where the user no longer has access to the calendar.

Those should be deleted automatically. There are sharing hooks for it.

jancborchardt:
Yes, that should all be automatic. No additional settings for users or admins for this needed. :) @ben-denham can you change that?

ben-denham:
Ok, calendar-user preferences are now deleted automatically when:
- The calendar is deleted (unchanged in latest commits).
- The user is deleted (unchanged in latest commits).
- The calendar is unshared with the user.
- The calendar is unshared with a group the user is a member of.
- A group is deleted that the calendar is shared with and the user is a member of.
- The user is removed from a group that the calendar is shared with.

However, preferences are only deleted if the user does not have access to the calendar in some other way (e.g. being the owner or a having a second share) :)

jancborchardt:
Cool, so if now there’s no interface change anymore I can at least give my design-:+1: :)

georgehrke:
@butonic What was the limit for table names in oracle? 26 chars?

butonic:
@georgehrke 30 chars

jancborchardt:
@georgehrke so what’s your call on this?

@rysiekpl can you review this as well?

tanghus:
@ben-denham Sorry for the long delay. Can I get you to rebase to master?

ben-denham:
@tanghus Just rebased to master.

tanghus:
Thanks @ben-denham :)

lib/calendar.php
```diff
@@ -249,7 +397,7 @@ public static function touchCalendar($id) {
 	 */
 	public static function deleteCalendar($id) {
 		$calendar = self::find($id);
-		if ($calendar['userid'] != OCP\User::getUser() && !OC_Group::inGroup(OCP\User::getUser(), 'admin')) {
```
tanghus, line 252: Removing the admin check means you'll get an error when admin deletes a user possibly causing other cleanup hooks to fail. Since `OC_User::isAdminUser(OC_User::getUser())` is not in the public namespace, this is the only way to check.

lib/calendar.php
```diff
@@ -283,7 +434,7 @@ public static function deleteCalendar($id) {
 	 */
 	public static function mergeCalendar($id1, $id2) {
 		$calendar = self::find($id1);
-		if ($calendar['userid'] != OCP\User::getUser() && !OC_Group::inGroup(OCP\User::getUser(), 'admin')) {
```
tanghus, line 286: Same here

lib/hooks.php
```diff
@@ -35,6 +35,98 @@ public static function deleteUser($parameters) {
...
+					OC_Calendar_Calendar::setCalendarDefaultUserPreferences($user, $parameters['itemSource']);
+				}
+			}
+		}
+		return true;
+	}
+
+	/**
+	 * @brief Delete obsolete user preferences after a calendar share is deleted.
+	 * @param paramters parameters from post_unshare-Hook
```
tanghus, line 67: typo ;)

lib/hooks.php
```diff
@@ -35,6 +35,98 @@ public static function deleteUser($parameters) {
...
+
+	/**
+	 * @brief Add initial calendar user preferences for calendars
+	 *        shared with a group when a user is added to the group.
+	 * @param paramters parameters from post_addToGroup-Hook
+	 * @return array
+	 */
+	public static function addToGroup($parameters) {
+		// Get all calendar ids shared with the group.
+		$stmt = OCP\DB::prepare( 'SELECT `item_source` FROM `*PREFIX*share` WHERE `item_type`=? AND `share_type`=? AND `share_with`=?' );
```
tanghus, line 91: **Never** touch `oc_share` table directly from an app!

lib/hooks.php
```diff
@@ -35,6 +35,98 @@ public static function deleteUser($parameters) {
...
+
+	/**
+	 * @brief Delete obsolete calendar user preferences when a
+	 *        user is removed from a group.
+	 * @param paramters parameters from post_removeFromGroup-Hook
+	 * @return array
+	 */
+	public static function postRemoveFromGroup($parameters) {
+		// Get all calendar ids shared with the group.
+		$stmt = OCP\DB::prepare( 'SELECT `item_source` FROM `*PREFIX*share` WHERE `item_type`=? AND `share_type`=? AND `share_with`=?' );
```
tanghus, line 108: Same here

lib/hooks.php
```diff
@@ -35,6 +35,98 @@ public static function deleteUser($parameters) {
...
+	/**
+	 * @brief Delete obsolete calendar shares when a group is deleted.
+	 * @param paramters parameters from post_deleteGroup-Hook
+	 * @return array
+	 */
+	public static function postDeleteGroup($parameters) {
+		// Get users in group. Will this work since the group is already deleted?
+		$users = \OC_Group::usersInGroup($parameters['gid']);
+		// Get all calendar ids shared with the group.
+		$stmt = OCP\DB::prepare( 'SELECT `item_source` FROM `*PREFIX*share` WHERE `item_type`=? AND `share_type`=? AND `share_with`=?' );
```
tanghus, line 125: And here.

tanghus:
@ben-denham Generally the code looks good, but the direct manipulation of the share table is a no-go.

Instead use something like this:

Add a method in lib/hooks.php:

``` php
    public static function calendarDeletion($parameters) {
        // Clean up sharing
        \OCP\Share::unshareAll('calendar', $parameters['id']);
}
```

And in appinfo/app.php add:

``` php
OCP\Util::connectHook('OC_Calendar', 'deleteCalendar', 'OC_Calendar_Hooks', 'calendarDeletion');
```

lib/calendar.php
tanghus, line 252: Just noticed that I used non-public `OC_Group` here. Old code ;)

vgezer:
Any updates on this? I really would like to see this on OC7 :)

jancborchardt:
@ben-denham yeah, me too. :) Can you rebase this branch?

lib/calendar.php
ben-denham, line 252: Ok I'll put this back.

Do you know what the public equivalent of OC_Group is? I can't find anything like it in core's lib/public

ben-denham:
@tanghus Ok, I have fixed the typos, and I have stopped using the oc_share table directly.

However, when using the public share API, we cannot get a list of calendars that are shared with a particular group, which has caused me to make two key changes:
- When a user is removed from a group, or a group is deleted, we must check all calendars that a user has preferences for to see if any preferences need to be removed (i.e. The user no longer have permission to access the calendar), rather than just checking the calendars that were shared with the group.
- When a user is added to a group, there is no way to know what new calendars may be shared with them through the group, so we do not know which calendars to add default preferences for. Therefore, I have changed the logic so that whenever a calendar is being retrieved (through allCalendars() or find()), we will add default preferences if no preferences currently exist. The default preferences will only be added the first time the calendar is retrieved for the user.
  - Because we now add default preferences this way, I have removed the query that would add   preferences when the calendar app is updated.

Also, as you say that OC_Group is private and should not be used, do you know the correct way to check if a user is an admin? It seems that OCP\User::checkAdminUser() would redirect the user to the home page if they were not an admin.

tanghus:
@georgehrke can you take over from here?

lib/calendar.php
```diff
@@ -52,10 +45,19 @@ public static function allCalendars($uid, $active=false) {
 				| OCP\PERMISSION_READ | OCP\PERMISSION_UPDATE
 				| OCP\PERMISSION_DELETE | OCP\PERMISSION_SHARE;
 			$row['description'] = '';
-			$calendars[] = $row;
```
rwasserstrom, line 54: I am looking at my owncloud files (version 0.6.3) and I don't see a lot of these lines of code anywhere in the files. Is there a reason why I see something different from what you have? Thanks

rwasserstrom:
I am looking in our files and we don't have the same code for a lot of these calendar files. Our owncloud is 6.3, fresh install. Please advise...

vgezer:
@rwasserstrom This is an unmerged branch. You will be seeing this feature when it is merged.

rwasserstrom:
Ok, so can we help implement this branch? We would like to help since we need this feature.

vgezer:
@rwasserstrom You can fork https://github.com/ben-denham/calendar and work on `shared-calendar-preferences` branch. You may want to talk to @ben-denham @georgehrke as they were working on this recently.

PR #289, opened 2014-01-06
Event sharing fixes

ben-denham:
A couple of fixes related to individually shared events:
- When retrieving a list of events that have been shared with a user: events that belong to calendars that the user can access are ignored. This prevents events from being shown twice on the calendar view.
- A new "Shared events" calendar is displayed in the list of calendars. This calendar allows users to hide/show all events that have been individually shared with them. This solves the problem of someone individually sharing 100 spam events with you, and you having no way to hide them.

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -448,6 +448,14 @@ Calendar={
 					}
 				  });
 			},
+			sharedEventsActivation:function(checkbox)
+			{
+				if (checkbox.checked){
+					$('#fullcalendar').fullCalendar('addEventSource', sharedEventSource);
+				}else{
+					$('#fullcalendar').fullCalendar('removeEventSource', sharedEventSource.url);
+				}
+			},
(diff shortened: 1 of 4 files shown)


jancborchardt:
@georgehrke can you check this out as well? Also @michaelzoidl take a look at this if you like. :)

jancborchardt:
@nezzi @rysiekpl can you review this? :) If everyone tests everyone elses pull requests, we can fix stuff much faster. :)

tanghus:
Tested and works :+1: 
@georgehrke shall I merge?

georgehrke:
If it works, go ahead and merge it!
Thanks for help with reviewing. I really appreciate it.

tanghus:
np

jancborchardt:
@tanghus you’re awesome! :) Thank you for reviewing the pulls.

tanghus:
:blush:

PR #321, opened 2014-01-29
remove possibility to overwrite on import

jancborchardt:
Every time I want to import an event, I get scared at the import form. This checkbox to potentially delete the whole calendar is just a bit too easy to click.

If you really want to overwrite the calendar, then you should delete it in the calendar app. No need for the dangerous »yeah sure destroy my data while doing other stuff« checkbox.

@georgehrke @jbtbnl please review. :) (There’s some other »overwrite« stuff in the code which is just at »0«, not sure if it should also be ripped out.)

### DIFF
--- a/js/loader.js
+++ b/js/loader.js
@@ -124,7 +124,6 @@ Calendar_Import={
 				$('#calendar_import_process').css('display', 'block');
 				$('#calendar_import_newcalendar').attr('readonly', 'readonly');
 				$('#calendar_import_calendar').attr('disabled', 'disabled');
-				$('#calendar_import_overwrite').attr('disabled', 'disabled');
 				Calendar_Import.Core.send();
 				window.setTimeout('Calendar_Import.Dialog.update()', 250);
 			}
@@ -159,7 +158,6 @@ Calendar_Import={
 				}
 			}else{
 				Calendar_Import.Store.method = 'old';
-				Calendar_Import.Store.overwrite = $('#calendar_import_overwrite').is(':checked') ? 1 : 0;
 			}
 			return true;
 		}
--- a/templates/part.import.php
+++ b/templates/part.import.php
@@ -44,8 +44,6 @@ $defaultcolors = OC_Calendar_Calendar::getCalendarColorOptions();
 			<!--<input id="calendar_import_generatename" type="button" class="button" value="<?php p($l->t('Take an available name!')); ?>"><br>-->
 			<div  id="calendar_import_mergewarning" class="hint"><?php p($l->t('A Calendar with this name already exists. If you continue anyhow, these calendars will be merged.')); ?></div>
 		</div>
-		<input type="checkbox" id="calendar_import_overwrite" value="1">
-		<label for="calendar_import_overwrite"><?php p($l->t('Remove all events from the selected calendar')); ?></label>
 		<br>
 		<input id="calendar_import_submit" type="button" class="button" value="&raquo; <?php p($l->t('Import')); ?> &raquo;" id="startimport">
 	<form>


jbtbnl:
It could also be hidden like "Advanced options" as seen when creating an appointment.

I'm not sure what the use-case for the function would be but sometimes it is easier to just empty your calendar if you are not interested in past appointments instead of deleting duplicates afterwards.

jancborchardt:
It’s a dialog with a single use – importing an event. We don’t need »advanced options« for a second option that probably no one is ever going to use.

jancborchardt:
I mean – importing an event to the Calendar through uploading the ICS to Files and then opening it is completely undiscoverable anyway …

jbtbnl:
You are right, the ICS upload function should be located in the Calendar app. Opening an ICS file through Files is just a nice gimmick.

knallio:
Before you remove that feature, I just want to add that I actually use this. The only way to use my work calendar at the moment is to upload the whole calendar (automatically) from outlook at work via webdav to owncloud. Then I open that calendar file in Files and replace the old one, using the delete checkbox. 
One suggestion: I could live with an "are you sure" popup. This way, deleting the calendar by accident should be impossible, and I have only one click more.

jancborchardt:
@knallio adding yet another popup will not solve things, but just add frustrations. People generally do not read popups but just click »ok«.

But interesting you mention that you need to use this workaround … why is it not possible to use CalDAV? Also, ownCloud is a calendar server, not really a client – so you would just not have the work calendar in the web interface – but instead on your mobile device and desktop.
Is your work calendar subscribable in some way, like via password-protected ICS?

Short: What can we do to fix your problem properly?

knallio:
As far as I can see, outlook does not support caldav. You need to pay for a 3rd-party add-in. So I was kind of happy that it was at least possible to upload the calendar via webdav.

knallio:
To answer your question: I am not sure what you can do to fix my problem, since it is a limitation in outlook. What might help is a way to define an "external" calendar source in the calendar app. which would point to the ics file. That way I can skip the whole import process and everything would be automatic. As far as I understood, this might be possible with the calendar rework?

craigarno:
> As far as I can see, outlook does not support caldav. You need to pay for a 3rd-party add-in.

Would something like this work?  http://sourceforge.net/projects/outlookcaldav/

Addendum: couldn't find anything to download.  i.e. only looks like a framework for the project, nothing deliverable yet.
Wonder if there are other "free" solutions available today?

Like http://support.atmail.com/display/A7DOCS/Outlook+CalDAV+Synchronization+Setup
Download CalDAV support:  http://download.atmail.com/getsync.php

knallio:
@craigarno Yes and no. In principle it would work, but I doubt that I can convince our IT department to install any add-in in that direction, especially an alpha version with (as far as I can see) very little activity. But thanks for the suggestion.

georgehrke:
@knallio Are you the admin of the ownCloud instance and can execute cron jobs on the server? Then you might be interested in https://github.com/georgehrke/cl-calendarimport

jbtbnl:
@georgehrke nice, I didn't know such a script did exist already. Why isn't this part of the calendar app? It shouldn't be too difficult to create some sort of scheduler that runs this periodically.

knallio:
@georgehrke That might be a great solution. I am the admin of the ownCloud instance. However, I cannot execute cron jobs, but I might find a solution. Some kind of integration would be cool, maybe as an additional app, making it configurable from within owncloud. Is there a way to hook into ownclouds cron system (AJAX) in case one cannot use the servers cron? 

I will definitely have a look at that script, thank you!

jancborchardt:
So – can we merge this pull request? The checkbox and possibility to lose my whole calendar creeps me out every single time I import an .ics

jbtbnl:
I believe it can be merged, but I'll leave that up to the calendar maintainers.

@knallio please open an issue for the integration of the import script with the calendar app by means of the owncloud cron system. It is a good idea!

georgehrke:
@jancborchardt Wouldn't simply reverting https://github.com/owncloud/calendar/commit/6ef96d21fc4397cb101567b63e3c9038e9202449 be easier?

georgehrke:
@jancborchardt  I've never used this feature at all, but you can see there are users who actually use this feature. Simply removing a feature that has been there for quite some time isn't really the way we should handle things like this.

jancborchardt:
@georgehrke well, I would say on the other hand »simply adding features which we don’t really know the use of isn’t really the way we should handle things«. ;)

But now I’m not really sure anymore if removing it is better than leaving it there anyway. It’s difficult enough to import an event so let’s maybe leave it.

PR #357, opened 2014-03-02
add auto-refresh to calender and controls

wtayyeb:
Dear OC Team.
There is no button on the web interface that refresh the calendar. then one have to refresh whole page(F5) or go one week later and come back to get calendar refreshed.
Very useful when working with calendar on smartphone(via DAV) and web Interface on PC.

I have make the patch and it is very clear and hope it will be useful.


georgehrke:
cc @jancborchardt

georgehrke:
if anything, the calendar should refetch events itself (like every 5 minutes). I don't see a benefit from a refetch button over just reloading the page. It adds just yet another button.

georgehrke:
:-1: from my side, sorry

wtayyeb:
@georgehrke, thanks for reply. 
reloading the page will lead to loading html, js, css also.
and as you mention after 5min, who will wait for it?

OK, if you think it is usefulness button, maybe we can add it to settings part.

georgehrke:
> reloading the page will lead to loading html, js, css also.

It just reloads the html and the events. JS and css are cached.

> and as you mention after 5min, who will wait for it?

For sure. The best would be:
- Server notifies webbrowser  via websockets when something changes
- webbrowser refreshes event sources when the changes affect the timerange that is currently being displayed.

but that's something we can't do when we want to support all platforms.

> OK, if you think it is usefulness button, maybe we can add it to settings part.

Don't get me wrong :) but that would be even less useful, because you have to open the settings, which will load the settings dialog (html + js) from the server and then click refresh instead of just reloading the page

wtayyeb:
websockets is great. but needs to code it. and side-load just for things not that important for all.
Google calendar was adding it as a direct button in its previous design and now move it in dropdown button with Print feature. I dont mean to blind copy from google. No. I think it is useful while you think not.

In other way to  implement such this idea, but with more effort, we can add the corresponding js to  current buttons. mean when we in week view if click on week view button again it will refresh it. currently the buttons doesn't affect anything.(fullcalendar bypass it). and so on for other views.

georgehrke:
What's your opinion @jancborchardt ?

jancborchardt:
I am against a dedicated refresh button. The proper solution is what has already been mentioned by @wtayyeb and @georgehrke:
- Reload the events on every view-change action in the interface (day/week/month), and by pressing »Today«, and also when hiding or showing a calendar.
- Also reload the events in regular intervals, for example every 5 minutes.

The reload should be in the background. People should not even know that it was reloaded.

georgehrke:
@wtayyeb Could you also implement the automatic reload :)
It's just some window.setTimeout( ... , 300000) magic

wtayyeb:
@georgehrke the today button also do the trick and the 5 min interval also implemented.
I decide to make the auto refresh to be 5 min after last action. it is better than just solid 5 min interval.

jancborchardt:
Ok, so this pull request does not introduce any more interface elements, right?

wtayyeb:
@jancborchardt Yes, no interface element added. as you can see the JS file just edited.

jancborchardt:
@georgehrke you closed this without comment – did you merge it manually or what was the resolution on this?

georgehrke:
oh, sorry.
@jancborchardt thanks for notifying

Code looks good from my point.
@jancborchardt Do you give a :+1: ?

georgehrke:
ping @jancborchardt

jancborchardt:
Looks good to me :+1:

jancborchardt:
@wtayyeb thank you for your contribution! :)

Are you on our IRC channel #owncloud-dev already? We also have a developer mailing list: http://mailman.owncloud.org/mailman/listinfo/devel

wtayyeb:
@jancborchardt thank you for merge! :)
no I don't use neither IRC nor mailing-list. maybe in the future I join the mailing list only.

PR #308, opened 2014-01-15
[AWAITING REVIEW] Implementing #411 - public calendar sharing

rysiekpl:
The aim: implementing [public link-sharing of calendars](https://github.com/owncloud/apps/issues/411). The functionality that works:
- link-sharing a calendar
  - displaying the link-shared calendar, along with events, info and download URL
  - downloading the shared calendar with the provided URL
  - password-protecting a link-shared calendar and password prompt;
  - accessing password-protected public calendars;
  - date-limit tested.
- link-sharing a single event
  - sharing GUI in pure HTML/CSS, with as little JS as possible (using CSS transitions, etc)
  - password-protecting
  - date-limiting
  - displaying
  - downloading iCal.
- new CSS/HTML interface for sharing calendars and events with users/groups on the same server
- CSS-only tooltips
- general fixes
  - information on event being shared publicly via a link-shared calendar has been added to event edit interface
  - changing the timezone now reloads calendar data (both in user's view, and in link-shared public view)
  - event datetime data are now adjusted to timezone
  - link-shared view of events and calendars allows the user to change the timezone data is displayed in.

The public link-sharing of both calendars and events can now be considered complete. Events can be individually link-shared regardless of their PRIVATE/CONFIDENTIAL status, but they will not show up (CONFIDENTIAL) or will only show as busy time (PRIVATE) in link-shared calendars.

PR #359, opened 2014-03-03
[WIP] redesign calendar app

jbtbnl:
Work in progress on introducing a sidebar in the calendar app to provide simpler navigation through the users calendars. Also the datepicker has been restyled to improve accessibility and clean up the UI.

Refer to https://github.com/owncloud/calendar/issues/269 for earlier discussion about the new design. Let's continue here to iron out the details.

Reference screenshot:
untitled

Checklist:
- [x] https://github.com/owncloud/calendar/issues/332 https://github.com/owncloud/calendar/issues/77 the calendar color should be in the calendar list on the left. Ideally just make it be the background of the checkbox (see Google Calendar, where the color box is the checkbox)
- [x] the calendar grid doesn’t seem to extend all the way to the bottom? There’s like 10–15 pixels whitespace left to the bottom of the screenshot
- [x] https://github.com/owncloud/calendar/issues/274 speed up settings menu
- [x] The settings area should not use tables (https://github.com/owncloud/calendar/pull/359#issuecomment-36745338 https://github.com/owncloud/calendar/pull/359#issuecomment-36747695)
- [x] Settings content should not exceed settings menu (https://github.com/owncloud/calendar/pull/359#issuecomment-36745338)
- [x] Rename a "hidden" calendar shows the calendar appointments after saving it (https://github.com/owncloud/calendar/pull/359#issuecomment-36745338)
- [x] Fix Firefox/Linux rendering (https://github.com/owncloud/calendar/pull/359#issuecomment-36746197 https://github.com/owncloud/calendar/pull/359#issuecomment-36747161)
- [x] Fix transparent checkbox color for some calendars (https://github.com/owncloud/calendar/pull/359#issuecomment-37561609)
- [x] Fix share icon for shared calendars (https://github.com/owncloud/calendar/pull/359#issuecomment-37561609)


jbtbnl:
CC @owncloud/designers

jancborchardt:
@jbtbnl dude, you are awesome! :) I have yet to try it out. But already some remarks (partly from #269):
- [x] https://github.com/owncloud/calendar/issues/332 https://github.com/owncloud/calendar/issues/77 the calendar color should be in the calendar list on the left. Ideally just make it be the background of the checkbox (see Google Calendar, where the color box is the checkbox)
- [x] the calendar grid doesn’t seem to extend all the way to the bottom? There’s like 10–15 pixels whitespace left to the bottom of the screenshot

georgehrke:
Wow, that looks fantastic! =D Thanks @jbtbnl !
Could you maybe take a look at the resize issue. The calendar and the left sidebar don't resize when resizing the browser window.

jbtbnl:
@jancborchardt thanks, it must be in the name ;)

> the calendar grid doesn’t seem to extend all the way to the bottom? There’s like 10–15 pixels whitespace left to the bottom of the screenshot

This occurs also in calendar/master, I tried to fix it but it seems that a lot of heights and widths are calculated by JavaScript and I've found no good entry point to start fixing it. (also @georgehrke)

wtayyeb:
Great, I have test it, but if there is a button to hide/show the sidebar it will be great.
It is owncloud design bug also. the owncloud not properly shown on smartphones.

its look like there is a bug there in the design
Firefox 27.0.1, Windows 7
screenshot 1393876049

jbtbnl:
@wtayyeb thanks for reporting, I will take a look at it. Hiding the sidebar is something that should be implemented in core, since it is a design element that is common to a lot of apps. Would you like to check if such an issue already exists in core? Look for mobile theme related issues.

wtayyeb:
there is an app named [MobileCompTheme](http://apps.owncloud.com/content/show.php/Owncloud+Mobile?content=163448) which just use css @media.

jancborchardt:
Yes, the mobile stuff should be separately done in core. We’ll probably work on that for ownCloud 7.

jbtbnl:
@jancborchardt this branch is currently at the point where we should decide on how to clean up the bunch of action icons at the right of each calendar name. In https://github.com/owncloud/calendar/issues/269 we talked about expanding a calendar entry downwards.

If we would like to implement this it would be best to implement this as a general design element in core. Do you agree? I do not yet have a good idea in mind on how such thing should be designed. Shall I create an issue in core or design so that we can discuss this with @owncloud/designers?

jbtbnl:
@wtayyeb the issues you experienced earlier should be fixed by https://github.com/owncloud/calendar/commit/29d796fcff984850a71badbbd708ea88a4b86845. Can you confirm this? Thanks!

jancborchardt:
@jbtbnl I would say we should do that in a separate discussion / pull request, otherwise this one grows too large. Is that ok? And yeah, I would also say it should be a general design element in core – great point! :)

jbtbnl:
@jancborchardt @georgehrke @raghunayyar please review this pull request, I feel like it's ready to merge this into master.

jancborchardt:
Yup, will do. Please wait for my review before merging. :)

raghunayyar:
@jbtbnl add outline:0; to the button styles. Plus most of the buttons are slightly misaligned in my machine. So, some CSS stuff is missing.

jbtbnl:
@raghunayyar what browser and OS are you using? A screenshot would also be helpful.
I will add outline:0; to the buttons, thanks.

raghunayyar:
@jbtbnl shouldn't clicking outside the side app-navigation calendar close it? cc @jancborchardt

jancborchardt:
@jbtbnl @raghunayyar the outline:0 on the buttons should probably be patched in core?

@raghunayyar no, the sidebar is permanent. Just like in News and Notes and Contacts.

raghunayyar:
@jancborchardt I will do a PR :)

jbtbnl:
@jancborchardt I think he meant the datepicker in app-navigation. In my opinion it is best to not close it automatically because one might need to utilize it multiple times.

raghunayyar:
^ correct :)

MorrisJobke:
@jbtbnl I have some design issues:
- [x] Setting table cells shouldn't highlight on hover (see picture)
- [x] Overlapping button (see picture)
- [x] Rename a "hidden" calendar shows the calendar elements after the click on the checkmark
- ~~Calendar actions (Rename, share, delete, ...) should appear on hover~~

calendar-rework

I'm irritated on the world button. It's weird that just a box appears ... but what is it. cc @jancborchardt Maybe need some UX improvement.

MorrisJobke:
@jbtbnl Otherwise: Thanks! It's much better than the previous way :)

raghunayyar:
@jbtbnl when I create a new calendar, the new entry gets in the view much lower than the list. Check the screenshot.
screen shot 2014-03-05 at 7 44 05 pm

jancborchardt:
Review:
- [x] the settings area is waaaaay too big! :) See https://github.com/owncloud/core/pull/7567 for fix
- [ ] can we please finally remove the »clear cache for repeating events«? https://github.com/owncloud/calendar/pull/252
- [ ] the share dialog doesn’t work as it stays the same height.

@MorrisJobke we shouldn’t do the hover-stuff for actions again – that does not work on tablets. cc @jbtbnl – _but_ the actions should be a bit greyed out, like half-transparent. Like in the files app. That should be achieved by attaching the .action class to it.

jancborchardt:
Otherwise super-awesome @jbtbnl :)

MorrisJobke:
@jancborchardt I'm fine with the greyed out stuff.

@jbtbnl Maybe have a look at the firefox rendering: 

calendar-rework-firefox

The text in the buttons is a bit misaligned and the "New calendar" is behind the week view. (shifted to the right - see scroll bar)

jbtbnl:
> the share dialog doesn’t work as it stays the same height

@jancborchardt I Guess we should address this with https://github.com/owncloud/core/issues/7563

> Maybe have a look at the firefox rendering

@MorrisJobke can you confirm that the button font is the Ubuntu font? In causes the buttons to be stretched too wide. Not sure if this should be handled here or in core. 

> when I create a new calendar, the new entry gets in the view much lower than the list

@raghunayyar thanks, will fix that. Either preliminary or by https://github.com/owncloud/core/issues/7563 as the space in between is occupied by the CalDAV link dialog.

MorrisJobke:
> @MorrisJobke can you confirm that the button font is the Ubuntu font? In causes the buttons to be stretched too wide. Not sure if this should be handled here or in core.

It's Cantarell. But if I remove the `padding: 8px` from `#navigation-list input[type="button"]` all is fine in Firefox. (also the scrollbar issue and the moved out "New calendar" https://github.com/owncloud/calendar/pull/359/files#diff-6a6a912d0c7b55b537768da778032849R300

jancborchardt:
@jbtbnl no, that’s a different thing. The share dialog should expand properly, and it does in current master. Can’t see what the problem is though.

Btw, pulling in @rysiekpl for review. Also, this probably clashes a bit with his pull request for sharing #308 – so it would be cool if we can work on that as well. :)

jancborchardt:
Oh and another point:
- [x] the settings area should not use tables! :) That will also solve the color-change on hover problem.

MorrisJobke:
@jancborchardt What about using `white-space: normal;` for buttons? Then the too long button simply gets a 2-line-button and looks a lot better.

jbtbnl:
@jancborchardt can you take a look at the screenshot of @MorrisJobke https://github.com/owncloud/calendar/pull/359#issuecomment-36746197 because I think that this is a font issue that might originate from core, i.e. that there is no font-family set for buttons.

jancborchardt:
@MorrisJobke buttons should never be that long and never be wrapped. The true origin of the problem is the long label, or better yet the function itself. See #252
As for the screenshot in the comment about the font issue: Does that happen with all the buttons, also those in Files?

MorrisJobke:
> As for the screenshot in the comment about the font issue: Does that happen with all the buttons, also those in Files?

No. That's the cause of the button style of the calender (it's also changed in this PR)

jancborchardt:
@jbtbnl can you go through the thread and copy those issues which are still left up to the original post as a checkbox list? :)

raghunayyar:
@jbtbnl the app-navigation has a custom width of 250px in this case which is much lesser as compared to how it is in News, Notes. I guess we should use the same width to maintain consistency. cc @jancborchardt

jancborchardt:
The other question would be: Should the navigation be narrower in general? What do you think concerning Notes and News @Raydiation @cosenal and Contacts @tanghus?

BernhardPosselt:
I think the current calendar u interface is better. Trying to force a massive sidebar that only does navigation for 3 basic things (go to previous/today/next) is not a good idea, that's why you're having issues with the width.

georgehrke:
> I think the current calendar u interface is better. Trying to force a massive sidebar that only does navigation for 3 basic things (go to previous/today/next) is not a good idea, that's why you're having issues with the width.

It's not only about moving navigation but also about integrating the calendar list in the main view

BernhardPosselt:
Having a familiar pattern for every app is a good idea but you shouldn't do it if it obviously doesn't fit the main workflow/function

BernhardPosselt:
Should not be visible since you don't change your calendar a lot. Normally you either need one calendar or a view with all events combined. Compared to the news/notes app: changing notes/feeds is needed a lot in the normal workflow

georgehrke:
> Should not be visible since you don't change your calendar a lot.

I maybe don't edit my calendars that often, but I often change their visibility.

georgehrke:
> but I often change their visibility.

Which, right now is a pain, because you have to go to settings, just to show/hide a calendar

BernhardPosselt:
I suppose most people only have one calendar. If you only have one calendar its not needed to show the bar at all since there's nothing to switch. I'd make the calendar list hidden by default, toggling it should remember the state though. The main focus should be the calendar itself

raghunayyar:
I would rather prefer a higher width of the sidebar in calendar as it as at least more clickable areas than any other app around.

tanghus:
> Should the navigation be narrower in general? 

@jbtbnl has made it change size via media queries in contacts (and maybe also here?). Works quite well. Normal desktop monitors are more than wide enough for a wide side bar.

libasys:
My 5 Cents:
1. Left sidebar is useful if this is to show & hide (Think to tablet)
2. Listed Calendars should show their permissions per tooltip (my SharedCalendars permissions, Shared to me permissions)
3. Color of each Calendar for overview
4. De-/Activate Calendars options
5. A Mini Calendar for fast switching months

Summary:
I am not a fan for direct edit options on the left sidebar, cause how often you rename a calendar or the change the color? More js events! What I'am doing if i have more than the standard 3 cal switching modes (month,day,week)?

Here an other example for left side bar:
screen_left_nav

jancborchardt:
Well, we planned the sidebar for the calendar long already and have a consensus that it’s needed. When you have multiple calendars, the current interface is just insufficient to work with them.

Also look at different other calendar apps like Google Calendar, iCloud Calendar, etc etc – all have the sidebar with calendars.
Also, since screens tend to be more wide than high, getting rid of the top bar is of more value than the sidebar makes us lose space.

cosenal:
@jancborchardt 

> Should the navigation be narrower in general? (in Notes and News, for example)

No, in my opinion, because of the argument you just used in the previous comment:

> screens tend to be more wide than high

jbtbnl:
I've extracted several issues mentioned above and listed them here: https://github.com/owncloud/calendar/pull/359#issue-28614504
If you find that I missed something then please tell me :)

If there is a need for further discussion about design decisions  then please continue at #269 so that we don't mix it up with the discussion/issues of this pull request. Thank you!

jbtbnl:
@MorrisJobke can you please check if your Firefox render issues have been fixed?

wtayyeb:
@jbtbnl yes it is solved

jbtbnl:
@owncloud/designers @jancborchardt @georgehrke please review the latest state of this branch. We should decide whether it is ready for a merge or not.

js/calendar.js
```diff
@@ -464,16 +470,40 @@ Calendar={
 				}
 			},
 			newCalendar:function(object){
-				var tr = $(document.createElement('tr'))
+				var div = $(document.createElement('div'))
```
tanghus: `var div = $('<div />');`
I know, not your code ;)

templates/calendar.php
```diff
@@ -1,24 +1,116 @@
...
+					<label class="bold"><?php p($l->t('iOS/OS X CalDAV address')); ?></label>
+					<input type="text" value="<?php print_unescaped(OCP\Util::linkToRemote('caldav')); ?>principals/<?php p(OCP\USER::getUser()); ?>/" style="width:90%" readonly>
+				</li>
+			</ul>
+		</div>
+	</div>
+	<!-- End of settings -->
 </div>
-<div id="fullcalendar"></div>
+<div id="fullcalendar"><img id="loading" src="<?php print_unescaped(OCP\Util::imagePath('calendar', 'loading.gif')); ?>" /></div>
```
tanghus: `<div id="fullcalendar" class="loading">`

tanghus:
calendar_double_share

``` html
<a class="share permanent" title="Share Calendar" data-possible-permissions="31" data-item="2" data-item-type="calendar" href="#" style="background: url("/core/img/actions/shared.svg") no-repeat scroll center center transparent;">
<img class="svg" src="/core/img/actions/shared.svg">
</a>
```

`class="share" gives the icon already, so it's actually defined 3 times here.
- The activation checkboxes don't work as one would expect. It looks like you have to check them, but when you do they get greyed out and are actually deactivated.
- One of the calendars are shared. That should probably be reflected in the display name.
- If you click the share icon and - if you haven't noticed the input has focus - clicks in the input, the dialog slides up again.
- When changing "Week starts on..." the fullcalendar should maybe reload. Not sure if it's at all related to this PR?
- I don't know if the undefined `itemSourceName` is related to this either?
  calendar_event_share_undefined

tanghus:
Otherwise it's supa-slick :+1: :rocket:

js/calendar.js
jbtbnl: Thanks :)

templates/calendar.php
jbtbnl: That looks a bit silly because the borders of the calendar table are on top of it. 
untitled
Any ideas on how to fix that?

tanghus: I didn't test it, so I actually don't know where it's placed :smile: It was mostly to be able to get rid of duplicate images in core and calendar and be sure to use the right one in case core changes it.
I guess you could do something like:
`<div id="fullcalendar"><span id="loading" class="loading" />` Not sure if it should be span or div, and you would probably need to set height/width on it.

jbtbnl:
@tanghus I'm unable to replicate your share-icon issue, the generated source on my end is:

```
<a href="#" class="share permanent" data-item-type="calendar" data-item="1" data-possible-permissions="31" title="Share Calendar">
<img class="svg" src="/owncloud/core/img/actions/shared.svg">
</a>
```

So without the background declaration in the a element. Also I can't find an occurance of `class="share"` that declares a background icon.

**Update:** I can reproduce this now, the css is added to the share action element when a calendar is shared. Strange behavior...

jbtbnl:
> One of the calendars are shared. That should probably be reflected in the display name.

@jancborchardt any thoughts on this?

> When changing "Week starts on..." the fullcalendar should maybe reload. Not sure if it's at all related to this PR?

@tanghus Also occurs in stable 6, please open a separate issue for that.

> The activation checkboxes don't work as one would expect. It looks like you have to check them, but when you do they get greyed out and are actually deactivated.

The problem seems to be that not all calendars are assigned a color - causing the checkbox to be transparent while enabled. @georgehrke should there be a default color in CSS or is it a bug that some calendars (default one also) don't have a color assigned?

tanghus:
> Also occurs in stable 6, please open a separate issue for that.

Done.

> I can reproduce this now, the css is added to the share action element when a calendar is shared. Strange behavior...

Yes, it's done by the JS for sharing. It should change the class from `share` to `shared` instead.

templates/calendar.php
jancborchardt: It needs to be a separate element, so it can be removed/hidden again as well as being overlaid on top of fullcalendar.
`<div id="fullcalendar"><div id="loading" class="icon-loading"></div></div>`

jancborchardt:
> One of the calendars are shared. That should probably be reflected in the display name.

No, it should be reflected via the permanent share icon. By default, the share icon should be this angle-thing with dots (share.png) – then when it’s shared with people, it should be shared.png, and when it’s shared publicly it should be public.png
Just like in the Files app.

jbtbnl:
@jancborchardt any idea why `display:none` is defined in [apps.css#L101](https://github.com/owncloud/core/blob/master/core/css/apps.css#L101) for recursive lists?

I am trying to fix the share dialog but the `shareWithList` is hidden by the above mentioned...

**update:** in https://github.com/owncloud/calendar/commit/5250743efe5f473e6be0321dea87c8d62cbc302d I overruled a couple of things from apps.css as a quick fix. A permanent solution should be discussed at https://github.com/owncloud/core/issues/7563

jbtbnl:
@owncloud/owners @tanghus @georgehrke please review the latest state of this branch :)

tanghus:
- The "checkboxes" looks awesome, although maybe not very standard UI element? I can disable my own calendars, but not the shared one. Is that intended, or is it not in the scope of this PR?
- There's still the - very minor - issue I mentioned earlier: If you click the share icon and - if you haven't noticed the input has focus - clicks in the input, the dialog slides up again.

libasys:
@tanghus the shared calendar can't disabled at the moment. 
A little workaround:
Save the active/deactive of a shared calendar in the personal settings of a user and check this when the eventsource get loaded! (js/l10n.php)

tanghus:
> the shared calendar can't disabled at the moment. 

@libasys OK, just wanted to make sure it wasn't related to this PR.

jbtbnl:
> There's still the - very minor - issue I mentioned earlier: If you click the share icon and - if you haven't noticed the input has focus - clicks in the input, the dialog slides up again

@tanghus it's a minor issue that in my opinion is not worth fixing right now because it will become obsolete anyway when https://github.com/owncloud/core/issues/7563 is implemented

> The "checkboxes" looks awesome, although maybe not very standard UI element? I can disable my own calendars, but not the shared one. Is that intended, or is it not in the scope of this PR?

There is an issue for that: https://github.com/owncloud/calendar/issues/21

tanghus:
:+1: :dancers: :palm_tree: :cocktail:

MorrisJobke:
I like it! :+1: from me, but I think I didn't test every possible case. But maybe we try to get this in and refine it where this is needed. For example the slightly greyed out calendar actions. @tanghus You're the master of this repo.

@jbtbnl You have done great work!

MorrisJobke:
Sorry. Just mixed up @georgehrke with @tanghus ... @georgehrke should have a look at this.

georgehrke:
I'll give it a final review tomorrow

raghunayyar:
This looks awesome to me as well. Great job once again. :)

tanghus:
> I'll give it a final review tomorrow 

@georgehrke ^

I'm afraid there will be conflicts with #308 though, so you have to decide which goes first.

libasys:
screen_redesign

some questions: i have made some fixes and updated the redesign tree to fullcalendar 1.6.4 js. how can i update/merge this now to [wip] redesign calender?
This is the repo: https://github.com/libasys/redesigncal602, only the changed files uploaded

jbtbnl:
@libasys it seems that you created a new repository, with only the files that have been changed in it. This will not merge upstream because they are detached from each other.

What you could do is the following:
- clone the calendar/redesign branch
- include your changes
- commit your changes
- push to a new private repository
- create a pull request for calendar/redesign

Then GitHub will come up with a nice comparison view showing what lines have been changed and this will enable the developers around here to verify it and merge it into calendar/redesign and later on calendar/master.

libasys:
Have the link to contacts/redesign branch? why contacts/redesign, this is calendar!

jbtbnl:
@libasys ehm, calendar/redesign offcourse, my bad :smile: 

Just clone https://github.com/owncloud/calendar.git and then switch to redesign branch.

libasys:
somebody works with aptana studio and team git, to help me?

jbtbnl:
@georgehrke I'd like to merge this before it becomes outdated please :)

georgehrke:
My last examination for this semester is tomorrow. I'll review tomorrow afternoon.

jbtbnl:
@georgehrke OK great, I'm sorry for bothering you then!

georgehrke:
:+1::+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: 

Sooo awesome!! thanks a lot :D

georgehrke:
one little thing: automatic resizing of calendar doesn't work on y-axis

libasys:
@georgehrke  No Prob! a big thanks back to the dev ;)

jbtbnl:
@georgehrke I'll create a separate issue for the vertical resizing because it's not a regression and not really related to the redesign.

jancborchardt:
WOW! This is finally in! So great, thank you @jbtbnl

jbtbnl:
@jancborchardt thanks, it took three months to get this in... worth the effort though :)

PR #376, opened 2014-03-20
Simple feature to configure alarm/reminders for an event

Iridias:
This patchset allows one to configure a simple alarm/reminder for an event in the calendar.
A select-box allows to choose pre-configured values for the duration. It will be embedded as an VALARM in the event-data, so that clients, which support alarms, will be able to remind you of an event, if you sync with them.
## Known issues:
- this will NOT enable owncloud to send any alarms/reminders! BUT if you sync your calendar with e.g. a smart-phone, the calendar in this client usually can handle the reminder! (At least BusinessCalendar can, and that was also my motivation to implement this feature)
- For the sake of a more easy implementation I decided to go with pre-configured values for the duration, although the format and most clients allow custom values! I used the durations that BusinessCalendar shows me when configuring an alarm, so it should be a good start.
- I also only support the alarm-action "DISPLAY", although the format also allows mail- and audio-actions
  - Also this feature only supports one alarm per event - haven't tried what happens, if you sync with a client where multiple alarms are configured


georgehrke:
I hope to find some time for reviewing this weekend. 
Just two little remarks:
- [ ] please generate this in a loop: https://github.com/owncloud/calendar/pull/376/files#diff-a7f0ad1a8a6c7d48b2d4fef693b2e4d5R62
- [ ] translations are handled over transifex, your changes to the language files will be overwritten, please contribute those translations on transifex

Iridias:
I can try to generate the duration-values in a loop, but I guess it won't be any shorter.
But thanks for the hint regarding the translations. I'll have a look how to do it with transifex.

georgehrke:
> I'll have a look how to do it with transifex.

Once merged, the strings inside a $l->t(''); will be indexed and uploaded to transifex

raimund-schluessler:
This seems to work nicely. Alarms can be set in the edit form and the sync also works (at least with thunderbird/lightning). Thanks for that!

I still have a small annotation. In https://github.com/owncloud/calendar/pull/376/files#diff-a944e41374cd4ba191f79e0cc12bd2e9R1059 it might be better to assign the value of the alarm-action by the request and set the default value "DISPLAY" in the user interface (it must not be shown though). This way a custom value can be set by other apps using the CalDAV backend (e.g. a task app, see https://github.com/raimund-schluessler/tasks_enhanced/issues/5).

With this small change I am for merging this PR.

PS: Something went wrong with the first comment I made here, it showed up twice, I deleted one, and both were gone. Sorry for that.

Iridias:
The initial requested/recommended improvements have been implemented in the pull-request a while ago. Any plans to merge it (or the enhanced version #410 from ElieSauveterre) anytime soon?

sysdev41:
Just downloaded 7.0.0.beta1 but this change seems not to be applied. When/in which version will this patch be applied?

sysdev41:
@Iridias : 
i am really looking forward to see this patch in owncloud. Just wrote to georg about merging this PR and got his feedback: 

" this pr has quite a few issues:
- it can handle only one alarm
- it always show the first alarm (doesn’t matter if it supports it (ACTION = DISPLAY) or not)
- iirc editing an event with multiple alarms will delete all but the first
- there are unresolved todos left from the dev
  Especially the deleting existing stuff is an absolute no-go and makes it a won’t-merge."

I can offer my support by testing some new releases of your PR.

georgehrke:
closing in favor of #410

PR #378, opened 2014-03-22
Add 4-Weeks view

citronalco:
Shows the current week at the top, followed by the next 3 weeks, like in Google Calendar

### DIFF
--- a/3rdparty/fullcalendar/js/fullcalendar.js
+++ b/3rdparty/fullcalendar/js/fullcalendar.js
@@ -45,11 +45,13 @@ var defaults = {
 	
 	// time formats
 	titleFormat: {
+		fourWeeks: 'MMMM yyyy',
 		month: 'MMMM yyyy',
 		week: "MMM d[ yyyy]{ '&#8212;'[ MMM] d yyyy}",
 		day: 'dddd, MMM d, yyyy'
 	},
 	columnFormat: {
+		fourWeeks: 'ddd',
 		month: 'ddd',
 		week: 'ddd M/d',
 		day: 'dddd M/d'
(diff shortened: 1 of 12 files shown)


georgehrke:
cc @jancborchardt

georgehrke:
4 weeks view should replace month review, no need to have both

citronalco:
disagreed. month view and 4 weeks view are used for different purposes. removing month view would irritate many users. including myself.

georgehrke:
there is no real benefit of month view, if there is a 4 week view

jancborchardt:
Totally agree with @georgehrke. The month view is useless if there is a proper 4 week view which puts the current week up top.

For reference see this post by Marco Arment: http://www.marco.org/2010/03/28/more-ideas-than-time-logarithmic-calendar-view

jancborchardt:
Even better of course would be if the whole thing would be continuous …: #360 ;)

cc @jbtbnl on this as well – what do you think?

citronalco:
I totally agree with Marco Arment when it's about the current month or the next month.
But I really prefer having the first day of the month on top when planning events far in the future, e.g. my christmas vacation next year.

jbtbnl:
@jancborchardt honestly I did not know that Google calendar has a 4-week view :smile: I use my calendar mainly in week view because of the amout of detail I've got in it. To make a new appointment in the future I  navigate to the right week via the datepicker.

Anyway, I am not against the fact that others want a 4-week view or whatever additional view. I might even end up using it! Just keep in mind that it will clutter the UI more.

jbtbnl:
@citronalco would you please rebase this so that we can test it with the new sidebar?

jancborchardt:
@citronalco can you rebase this? :) If you need help, just ask here.

citronalco:
See https://github.com/owncloud/calendar/pull/511
And yes, I really should improve my git knowledge.

PR #385, opened 2014-03-26
Allow only owner of calendar edit shared calendars

piesu:
Currently there are two problems:
- if anyone edits (changes name or color) shared calendar, then changes are propagated to all users
- there is bug that if anyone edits calendar, calendar becomes deactivated. It's pretty bad, as only owner can activate calendar back and it also affects all users with access to calendar

This workaround is far from best option (everybody can activate/deactivate and change color of only own copy) but prevents users from destroying something what they can't fix

### DIFF
--- a/lib/calendar.php
+++ b/lib/calendar.php
@@ -182,9 +182,6 @@ class OC_Calendar_Calendar{
 	public static function editCalendar($id,$name=null,$components=null,$timezone=null,$order=null,$color=null) {
 		// Need these ones for checking uri
 		$calendar = self::find($id);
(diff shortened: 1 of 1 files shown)


vgezer:
This PR could be related to this issue: https://github.com/owncloud/calendar/pull/288

georgehrke:
cc @tanghus

georgehrke:
merged manually to squash commits

master: https://github.com/owncloud/calendar/commit/e1effeb17e9ca43202f9908d126d04beff50385b
stable7: https://github.com/owncloud/calendar/commit/b434d2627a133de7342e374a8e8eace430bf1616

PR #389, opened 2014-04-03
Fix adding event dialog minor bug

wtayyeb:
Hi, the `Export event` button should not shown on new event dialog because export mechanism work with saved events in database. and currently when clicking on it, a 404 page displayed. So it must to be hidden on `Create new event` dialog.

Also I think it is good to trigger `Enter` button within event-title text-box so one can add event more quicker.

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -161,6 +161,11 @@ Calendar={
 			$('#advanced_month').change(function(){
 				Calendar.UI.repeat('month');
 			});
+			$('#event-title').bind('keydown', function(event){
+				if (event.which == 13){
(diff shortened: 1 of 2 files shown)


vgezer:
Tested and works :+1:. Thanks for commits.

cc @georgehrke 

Maybe we should add the same 'Enter' button for calendar adding and calendar editing.

vgezer:
@wtayyeb Do you want to create another branch and PR for calendar creation and editing with Enter key?

georgehrke:
code looks good, I'll test it later

PR #392, opened 2014-04-08
Fix untranslatable strings

vgezer:
- The default calendar name now can be translated
- Added a space between timezone and this will not affect
  existing translations


georgehrke:
these changes make sense. will test later today

PR #393, opened 2014-04-08
Fix uri encoding

vgezer:
Create a calendar with a name, e.g. ŞĞİüö (some special letters) and
try to get the CalDAV link. Accessing the link gave not found error.


vgezer:
@georgehrke Could you also test this? Thanks.

georgehrke:
thanks for your contribution :)

PR #394, opened 2014-04-09
Remove SanitizeHTML which encodes non-ascii characters while editing cal.

vgezer:
- Create a calendar with a non-ascii character: eg. üçöğ
- Edit the calendar with edit link
- The characters are encoded and not possible to see them properly.

@georgehrke What do you think?


vgezer:
@georgehrke Do we need to backport this also to OC6 and OC5?

georgehrke:
:-1: This will introduce XSS vulnerability.

vgezer:
Then how can we fix this problem?

georgehrke:
closing this in favor of https://github.com/owncloud/core/pull/8168

georgehrke:
double sanitization fixed in https://github.com/owncloud/calendar/commit/742048350a968676b8dae4d05186dd45442f5148

PR #399, opened 2014-04-09
Let users set a default calendar for new events

vgezer:
This should implement: #348

Users can set one calendar default by clicking on the star.
New events will be created in the default calendar.
- [x] If none is set, the first created calendar is the default one.
- [x] Setting a calendar as default highlights the star
- [x] Only one star is highlighted, remove the old one (needed to see the changes immediately)
- [x] Clicking on the same calendar should not do anything
- [x] New default calendar is used when clicked on calendar to create new event
- [x] The first calendar is used if user deletes the default calendar


vgezer:
Some screenshots:

_Test 2_ is selected as default:

calendar-new1

_Personal_ is selected as default:

calendar-new

georgehrke:
cc @jancborchardt

jancborchardt:
As this adds yet another control to the calendar rows, I’d say we need to solve first how we cope with 3+ control elements of that sort. Like introducing a caret/triangle to hint at more settings. And maybe have sharing always present (or whichever one we regard the most important action). And that should be done in core, in an apps.js (and existing apps.css).

@jbtbnl @Kondou-ger @MorrisJobke do you have proposals? Previously we talked about a bubble (not very well integrated), or expanding the row to expose options (well integrated as long as there is an animation). Anything else?

jbtbnl:
@jancborchardt I believe that a lot of "settings" do not actually belong there. For instance the two CalDAV links, they should be on the "Help" page instead. There could be a link in the settings drawer "Learn how to setup CalDAV" to guide new users there.

The "Clear cache" button shouldn't be necessary or should be at a less prominent location.

The "Timezone" setting is a general setting like display language and should therefore be in the personal settings page. On ownCloud setup it should default corresponding to the user's language.

georgehrke:
> The "Clear cache" button shouldn't be necessary or should be at a less prominent location.

What about showing the clear cache button only when debug mode is enabled?

> The "Timezone" setting is a general setting like display language and should therefore be in the personal settings page. On ownCloud setup it should default corresponding to the user's language.

We should get rid of the timezone setting anyhow. the calendar should just use the timezone of the browser.

jbtbnl:
> What about showing the clear cache button only when debug mode is enabled?

That could be, I can't judge if it's necessary that often while developing / debugging the Calendar app.

vgezer:
> @jancborchardt Like introducing a caret/triangle to hint at more settings.

I suggest using "..." as in Android or other mobile OSes.

ajax/calendar/makedefault.php
```diff
@@ -0,0 +1,26 @@
...
+ * later.
+ * See the COPYING-README file.
+ */
+
+
+OCP\JSON::checkLoggedIn();
+OCP\JSON::checkAppEnabled('calendar');
+OCP\JSON::callCheck();
+
+$cal = $_POST["calendarid"];
```
LukasReschke: isset?

ajax/calendar/makedefault.php
```diff
@@ -0,0 +1,26 @@
...
+try {
+	$del = OC_Calendar_Calendar::makeDefault($cal);
+	if($del == true) {
+		OCP\JSON::success();
+	}else{
+		OCP\JSON::error(array('error'=>'dberror'));
+	}
+} catch(Exception $e) {
+	OCP\JSON::error(array('message'=>$e->getMessage()));
+	exit;
```
LukasReschke: I guess the exit(); is not really needed.

ajax/calendar/makedefault.php
```diff
@@ -0,0 +1,26 @@
...
+
+OCP\JSON::checkLoggedIn();
+OCP\JSON::checkAppEnabled('calendar');
+OCP\JSON::callCheck();
+
+$cal = $_POST["calendarid"];
+
+try {
+	$del = OC_Calendar_Calendar::makeDefault($cal);
+	if($del == true) {
```
LukasReschke: Please use always the identical (`===`) operator, see http://www.php.net/manual/en/language.operators.comparison.php for more details.

ajax/calendar/makedefault.php
vgezer: @LukasReschke The exit was also in those files. That's why I did not modify actually: https://github.com/owncloud/calendar/blob/master/ajax/calendar/delete.php
https://github.com/owncloud/calendar/blob/master/ajax/import/import.php

ajax/calendar/makedefault.php
vgezer: Should I modify this as well:

https://github.com/owncloud/calendar/blob/master/ajax/calendar/delete.php#L14

LukasReschke: Go for it :-)

ajax/calendar/makedefault.php
LukasReschke: The one in delete.php seems uneeded. But the other ones in import.php are
not at the end of the file and I believe that they are needed :-)

Sent from mobile device

jancborchardt:
@jbtbnl:

> I believe that a lot of "settings" do not actually belong there. For instance the two CalDAV links, they should be on the "Help" page instead. There could be a link in the settings drawer "Learn how to setup CalDAV" to guide new users there.

No, all app-specific settings should be in the app. We previously had app settings on the personal page and it was confusing. As CalDAV is bound to the Calendar app, it should be in there. I agree that »Settings« might be a slightly misleading name, but I don’t think it’s a really grave issue.

> The "Clear cache" button shouldn't be necessary or should be at a less prominent location.

Yes, we should remove it for good. @georgehrke can we please do that? Just showing it in debug is also fine with me for now.

> The "Timezone" setting is a general setting like display language and should therefore be in the personal settings page. On ownCloud setup it should default corresponding to the user's language.

Yes. See the »global region format« issue in core: https://github.com/owncloud/core/issues/376

vgezer:
@LukasReschke Thanks :). I have committed again with the changes. Can you check it again?

georgehrke:
@jancborchardt What's your final decision on this?

I agree that yet another control element is not good. A better solution would be to make the calendar list sortable. Users could make a calendar 'default' by moving it to the top of the list.

vgezer:
@georgehrke Or maybe using the settings panel to pick the default calendar from dropdown list?

vgezer:
@jancborchardt, what should we do with this?

jancborchardt:
This is a good function, but we should wait with implementing it once we switch to the menu where items are extended downwards.

With the current solution of having icons next to it, we can’t just put function after function in there. Especially because in this case, the star icon is not very clear.

(Also, the ability to reorder calendars is a bit excessive to just get this solved, and is also not very discoverable that it defines the default calendar.)

So, we should wait with this. Hope that’s ok?

vgezer:
@jancborchardt Yes sure. I will update it later after new menus :)

brantje:
Rebased onto master

brantje:
:thumbsup: tested in chrome / ff / chrome mobile

brantje:
@jancborchardt why isn't this merged yet?

vgezer:
@brantje Because we wanted to avoid using extra action button. See the Jan's comment above: https://github.com/owncloud/calendar/pull/399#issuecomment-43312240

jancborchardt:
Yes, the sidebar is cluttered with actions at the moment. We need to clean that up before we add new stuff.

jbtbnl:
Work in progress on the action buttons cleanup: https://github.com/owncloud/core/pull/10644

vgezer:
What do we do with this? Should I work on this so this gets merged?

jbtbnl:
@wakeup since https://github.com/owncloud/core/pull/10644 has been merged, it would be great if you could add the actions menu in this PR.

@jancborchardt should all actions be in the menu or are there some (like edit and share) that should remain visible?

jancborchardt:
I’d say Share should remain visible, yes (just as planned for the Files app). It needs to be on half opacity though, and only be fully opaque when actually shared.

vgezer:
@jbtbnl Just a quick question, does the current git version of Calendar work in your environments?

Similar to https://github.com/owncloud/core/issues/12197 I am having this error when choosing Calendar app:

```
Fatal error: Uncaught exception 'Exception' with message 'js file not found: script:jquery.multiselect serverroot:/var/www/owncloud' in /var/www/owncloud/lib/private/template/resourcelocator.php:46 Stack trace: #0 /var/www/owncloud/lib/private/templatelayout.php(152): OC\Template\ResourceLocator->find(Array) #1 /var/www/owncloud/lib/private/templatelayout.php(103): OC_TemplateLayout::findJavascriptFiles(Array) #2 /var/www/owncloud/lib/private/template.php(118): OC_TemplateLayout->__construct('error', '') #3 /var/www/owncloud/lib/private/template/base.php(103): OC_Template->fetchPage() #4 /var/www/owncloud/lib/private/template.php(227): OC\Template\Base->printPage() #5 [internal function]: OC_Template::printExceptionErrorPage(Object(Exception)) #6 {main} thrown in /var/www/owncloud/lib/private/template/resourcelocator.php on line 46
```

jbtbnl:
@wakeup Calendar master doesn't work because of https://github.com/owncloud/core/pull/12005

georgehrke:
> @wakeup Calendar master doesn't work because of https://github.com/owncloud/core/pull/12005

should be fixed with https://github.com/owncloud/calendar/commit/59d249a5ced12a753e06e8b696485cc5926ca9a9

DeepDiver1975:
works :+1:

DeepDiver1975:
anyhow moving to backlog because of open UX issues as declared above

PR #401, opened 2014-04-10
More space between time title

vgezer:
On high res. monitors, it seems like there is almost no space between the time and title. This increases the gap between them and also adds a left padding to time.


georgehrke:
please create a new branch deriving from master branch. This derives from your default-calendar feature branch

PR #405, opened 2014-04-12
Select links when clicked

vgezer:
Users were needed to ~~double~~ triple click to select the whole addresses.

~~This makes the selection automatic like the calendars' CalDAV link.~~

This makes selection only clicking once instead of three.

### DIFF
--- a/js/on-event.js
+++ b/js/on-event.js
@@ -7,6 +7,12 @@ $('#caldav_url_close').live('click', function () {
 $('#caldav_url').live('mouseover', function () {
 	$('#caldav_url').select();
 });
+$('#primarycaldav').live('click', function () {
+	$('#primarycaldav').select();
+});
+$('#ioscaldav').live('click', function () {
+	$('#ioscaldav').select();
+});
 $('#editCategories').live('click', function () {
 	$(this).tipsy('hide');OC.Tags.edit('event');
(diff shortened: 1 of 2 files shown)


georgehrke:
But you actively click to get the caldav link, so automatic selection is the best there.
Primary calendar and principals urls are visible all the time, so mouseover to select could be a bit annoying and unexpected.

@jancborchardt What's your opinion?

vgezer:
Another thing can be clicking once to select the link. This case is more convenient.

jancborchardt:
Yeah, for a link which is displayed all the time, selection should take one click. But not happen on hover - keep in mind it needs to work on tablets too.

vgezer:
@georgehrke @jancborchardt Ok. Now they are selected when clicked only once. 

Can someone test it on tablets?

PS: I checked this on mobile. However, hovering on calendars' CalDAV link on mobile scrolls down the navigation_pane and makes it unable to copy the links. Can we modify it somehow?

jancborchardt:
Yup, works :+1:

@georgehrke @jbtbnl code review please?

georgehrke:
code looks good :+1:

georgehrke:
@wakeup thanks for your contribution :)

PR #421, opened 2014-05-04
Time and Title now have more space in between

vgezer:
@georgehrke A new commit only including that.

### DIFF
--- a/3rdparty/fullcalendar/css/fullcalendar.css
+++ b/3rdparty/fullcalendar/css/fullcalendar.css
@@ -287,7 +287,7 @@ a.fc-event {
 	
 .fc-event-time,
 .fc-event-title {
-	padding: 0 1px;
+	padding: 0 1px 0 3px;
      cursor: pointer;
 }
 


georgehrke:
cc @owncloud/designers

jbtbnl:
@wakeup please include a useful description because now I need to search for what has been fixed exactly.

jbtbnl:
Nonetheless :+1:

georgehrke:
@wakeup Would you mind contributing this fix to [upstream](https://github.com/arshaw/fullcalendar) as well?

vgezer:
Here is the before/after images:

web arayz1

web arayz1

When I first made a PR, it was more clear, but it seems this has already been solved so we can close this without merging.

@georgehrke Also we probably won't need to report into upstream. What do you think?

georgehrke:
> @georgehrke Also we probably won't need to report into upstream. What do you think?

If it was solved in the meantime and this needs no merge: yes

jbtbnl:
@wakeup what do you mean by that? In calendar/master the issue still exists. If it has been fixed in the latest FullCalendar version then we should merge that.

jbtbnl:
Also, for the second screenshot the second line of appointment "Calligra" should be indented as well.

jancborchardt:
This is something we should fix upstream only, not manually in ownCloud.

georgehrke:
see jan's comment

PR #432, opened 2014-05-30
replace all .live with .on

vgezer:
Mentioned here: https://github.com/owncloud/core/pull/8779#issuecomment-44651153

This replaces all .live() usage with .on()

Please test


georgehrke:
thx :)

PR #477, opened 2014-07-02
remove shadow effect from icon, reduce size

jancborchardt:
All shadow effects have been removed from core icons: https://github.com/owncloud/core/pull/9381

So the apps should follow. :) Please review @owncloud/designers @georgehrke


georgehrke:
:+1:

vgezer:
:+1:

PR #483, opened 2014-07-09
Timeline fixes

brantje:


js/calendar.js
```diff
@@ -116,9 +116,7 @@ Calendar={
 			var percentOfDay = curSeconds / 86400;
 			//24 * 60 * 60 = 86400, # of seconds in a day
 			var topLoc = Math.floor(parentDiv.height() * percentOfDay);
-           
-			timeline.css({'left':($(".fc-agenda-axis").width()+6)+'px','top':topLoc + 'px'});
-
+			timeline.css({'left':($('.fc-today').offset().left-250),'width': $('.fc-today').width(),'top':topLoc + 'px'});
```
PVince81: Is 250 the width of something specific ? Maybe you could read that value from the object directly.
If it's the sidebar, then use `$('.app-navigation').width()` instead (or `outerWidth()` if that works better)

PVince81:
@georgehrke please also review

PVince81:
Looks good :+1: 
Works in IE8, Firefox 29 and Chromium

PVince81:
Wait, there's still a bug: try resizing the window or zooming.

You might need to listen to resize and zoom events, then call your calculation function there.

brantje:
On it

libasys:
i am not sure if this works, cause fullcalendar.js also uses the window resize function!

brantje:
@libasys It works here, can you test?

libasys:
@brantje sorry i can't test it, i have a modified version of the calendar!

PVince81:
Thing is that the resize event can be caught/used by multiple handlers, as long as you use jQuery's `$(...).resize()`. If you use `window.onresize = handler` then it might not work and might overwrite whichever handler was registered before.

georgehrke:
:+1:

georgehrke:
@brantje thx for your contribution :)

georgehrke:
stable7: https://github.com/owncloud/calendar/commit/1445d6c7a71a5b13532518bb6cffbf7153ef7efd

Issue #463, opened 2014-06-26
Share box cannot be closed in Owncloud 7


maethor:
In Owncloud 7, the "share box" cannot be closed. When it is opened and I click on the share button, the share box closes and reopens directly. This is quite anoying because this box hides the following calendars in the list.

I'm using Firefox nightly 33.0a1 on Debian Wheezy and I have the same problem with Chromium 35.

Thanks.

jancborchardt:
This is most likely fixed. If it still occurs, please let us know with more details.

PR #487, opened 2014-07-13
Possible Fix #463

libasys:
Possible Fix for https://github.com/owncloud/calendar/issues/463
Please use the label only for the checkbox and not for the whole line, it runs into problems with the sharing dialog box!


georgehrke:
cc @owncloud/designers 

@libasys thx

vgezer:
Unfortunately, it does not solve the issue, but I think this is :+1:

MorrisJobke:
As @wakeup said: I doesn't fix the issue, but semantically its a better representation :+1:

georgehrke:
:+1:

georgehrke:
@libasys thx for the patch :)

georgehrke:
stable7: https://github.com/owncloud/calendar/commit/49e8185

PR #494, opened 2014-07-25
Fix activation of shared calendars.

michaelkuhn:
Editing shared calendars deactivates them and it is impossible to
activate them afterwards.

There are actually two problems:
- Shared calendars are deactivated when editing because they are
  missing their checkbox. Consequently, the checked status can not
  be read and the update function sets it to 0.
- As shared calendars do not have a checkbox, it is not possible to
  toggle their activation state afterwards.

This change adds a checkbox for shared calendars and allows their
activation, fixing both problems.

### DIFF
--- a/ajax/calendar/activation.php
+++ b/ajax/calendar/activation.php
@@ -12,7 +12,7 @@ OCP\JSON::checkAppEnabled('calendar');
 OCP\JSON::callCheck();
 
 $calendarid = $_POST['calendarid'];
-$calendar = OC_Calendar_App::getCalendar($calendarid, true);
+$calendar = OC_Calendar_App::getCalendar($calendarid, true, true);
 if(!$calendar) {
 	OCP\JSON::error(array('message'=>'permission denied'));
 	exit;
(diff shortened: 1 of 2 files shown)


georgehrke:
> This change adds a checkbox for shared calendars and allows their activation

This introduces a bug that allows all users to manipulate the activation state of a shared calendar which affects the owner and all user the calendar is shared with. This is why there is no checkbox for shared calendars,

georgehrke:
> Editing shared calendars deactivates them and it is impossible to
> activate them afterwards.

If I recall correctly this is caused by the update form always sending false for the enabled property. Hadn't had time to fix it yet

georgehrke:
[This commit](https://github.com/owncloud/calendar/commit/9ec3a26a0f43e9c1b478cba773de3fbf8598d6a2) fixes the bug that disables calendars on update, would be great if you could validate this fix :)

michaelkuhn:
Yes, the commit fixes the problem. Thanks!

Sorry, I was under the impression that other users should be able to activate/deactivate shared calendars, as they are also able to completely delete them.
I guess this should also be fixed then, right?

georgehrke:
> I guess this should also be fixed then, right?

Yes, it should unshare the calendar. I can look into it, a pull request would be appreciated though ;)

PR #495, opened 2014-07-26
Do not allow deleting shared calendars.

michaelkuhn:
This pull request makes sure that shared calendars can not be deleted but are instead unshared.

This change depends on owncloud/core#9915.

### DIFF
--- a/lib/calendar.php
+++ b/lib/calendar.php
@@ -254,6 +254,7 @@ class OC_Calendar_Calendar{
 	 */
 	public static function deleteCalendar($id) {
 		$calendar = self::find($id);
+		$isShared = ($calendar['userid'] !== OCP\User::getUser());
 		if (!self::isAllowedToDeleteCalendar($calendar)) {
 			$sharedCalendar = OCP\Share::getItemSharedWithBySource('calendar', $id);
 			if (!$sharedCalendar || !($sharedCalendar['permissions'] & OCP\PERMISSION_DELETE)) {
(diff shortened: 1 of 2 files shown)


michaelkuhn:
As owncloud/core#9915 has now been merged, please review this pull request again.

georgehrke:
Code looks good, still need to test it

michaelkuhn:
One question that came up while I was reading #555: deleteCalendar is also used to delete calendars when a user is deleted, correct? If so, I guess I will have to update the code to take that into consideration.

georgehrke:
yes, would be great if you could check

quenenni:
Hello,

I would like to know where you are with this problem? Will it be fixed soon?
(I'm using owncloud 8.0.6 and the problem still exists)

It's really an annoying feature/bug.. We have a multi instance owncloud and it happened several times that an user deleted shared calendars with all that involves (data loss, find backup, find the relevant tables / data to restore, ... ). 
That's very painfull ;)

georgehrke:
- [ ] needs rebase

zilon84:
Hello,

are there important reasons why this fix is not included, yet? I think it's very important to avoid data loss by an accidentally hit delete button by an user that does not own the calendar.

Thx,
Zilon

michaelkuhn:
Somehow I forgot to follow up on this, sorry about that.

I have updated the pull request with the workaround I currently use to prevent accidental calendar deletion (basically a rebased version of the original commit). I have also added a second commit to force calendar deletion during user deletion to address my earlier comment.

I am not sure whether this is sufficient, though. Someone with a better understanding of ownCloud's share system and the calendar app needs to review this (why do users of a shared calendar get the delete permission in the first place and would it be enough not to grant it?).

PR #499, opened 2014-07-27
Fixes #498

raimund-schluessler:
This PR fixes #498.


raimund-schluessler:
This and more is already included in #264. I will have a look at this PR instead. Maybe we can merge #264 instead. Sorry for the duplicates.

PR #511, opened 2014-07-31
Add 4-Weeks view

citronalco:
Shows the current week at the top, followed by the next 3 weeks, like in Google Calendar
only german translations included

This is the owncloud 7 version of https://github.com/owncloud/calendar/pull/378

### DIFF
--- a/3rdparty/fullcalendar/js/fullcalendar.js
+++ b/3rdparty/fullcalendar/js/fullcalendar.js
@@ -46,11 +46,13 @@ var defaults = {
 	// time formats
 	titleFormat: {
 		month: 'MMMM yyyy',
+		fourWeeks: 'MMMM yyyy',
 		week: "MMM d[ yyyy]{ '&#8212;'[ MMM] d yyyy}",
 		day: 'dddd, MMM d, yyyy'
 	},
 	columnFormat: {
 		month: 'ddd',
+		fourWeeks: 'ddd',
 		week: 'ddd M/d',
 		day: 'dddd M/d'
 	},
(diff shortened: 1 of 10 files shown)


georgehrke:
I still think we should remove month if we add 4-week

cc @jancborchardt

jancborchardt:
Yes, I agree. @citronalco can you remove the month view? The 4-weeks-view is better and makes the month view obsolete.

jancborchardt:
@citronalco also, please don’t commit any translation changes in the `l10n` folder. Translations are fully handled via [Transifex](https://www.transifex.com/projects/p/owncloud/).

jancborchardt:
Testing this it seems that without having continuous scrolling #360 there is an inherent problem here: When switching to months before or after, it’s also set to 4 weeks - no month overview or specific start and end date.

This really needs to be combined with continuous scrolling before we should merge it, let alone replace the month view with it.

jancborchardt:
@citronalco did you look at the comments? Can you make the changes? :) You’re also very welcome in our IRC channel #owncloud-calendar

dynek:
Hey, would it be acceptable to let the user actually selects how many days/weeks she/he wants to be displayed (Google Calendar-like: 2 Days, 3 Days, 4 Days, 5 Days, 6 Days, 7 Days, 2 Weeks, 3 Weeks, 4 Weeks). Not needed to go that far maybe (days) but 1 day + 1,2,3 weeks + 1 month is a good idea, isn't it ?

edit: it seems @citronalco left the conversation - I am interested in "taking over" this PR to go on with his work, can I go ahead ?

georgehrke:
> Hey, would it be acceptable to let the user actually selects how many days/weeks she/he wants to be displayed

It would be to complex to let the user select what exact view they want. Having options like google with 2 Days, 3 Days, 4 Days, 5 Days, 6 Days, 7 Days, etc ... is completely exorbitant.

We want to keep it simple and make users happy with good defaults.

dynek:
Reason why I said "not needed to go that far". It would be my first contribution to owncloud so I haven't dig too much into it yet. Would it be possible to create a dropdown list instead of the "Month" button and leave it to the user to select 1 week, 2 weeks, 3 weeks or 4 weeks ? Is it still too much ?
I'd love to have 3 weeks view.

georgehrke:
> Not needed to go that far maybe (days) but 1 day + 1,2,3 weeks + 1 month is a good idea, isn't it ?
> [...]
> I'd love to have 3 weeks view.

Why would you need 2 or 3 weeks?
What's the use case?

georgehrke:
My point is: We won't add a 3 weeks view just for the sake of having a 3 weeks view. There has to be a use case that benefits a remarkable share of our users.

dynek:
Well I'm being very selfish probably but I do use 3 weeks view since a long time on Google Calendar and find it very readable based on the number of entries I have daily. It's a good compromise between readability and being able to move items forward in the agenda.

Forget it - I'll apply the modification on my installation.

PR #527, opened 2014-08-13
Allow download of shared calendars and events

alevy:
The UI permits downloading a calendar shared with you, but the response
was Forbidden 403 becuase export.php wasn't setting the shared parameter
to true in it's calls to fetch the resource.


georgehrke:
:+1:

georgehrke:
thx for your contribution

georgehrke:
stable7: https://github.com/owncloud/calendar/commit/4d606bd

PR #559, opened 2014-09-14
Fixes the bug of sending an e-mail notification only to the owner of the...

nicolashainaux:
... event. Also allows to send e-mail notifications to the members of groups that are shared with. Also checks that each recipient gets the e-mail only once, even if it belongs to several groups, for instance.

### DIFF
--- a/lib/app.php
+++ b/lib/app.php
@@ -550,9 +550,8 @@ class OC_Calendar_App{
 				}
 			}
 			foreach ($eventsharees as $sharee) {
-				$eventShareesNames[] = $sharee['share_with'];
 				$shwth = $sharee['share_with'];
-                                if(OC_Group::groupExists($shwth)) {
+				if($sharee['share_type'] == \OCP\Share::SHARE_TYPE_GROUP) {
                                         foreach (OC_Group::usersInGroup($shwth) as $u) {
                                                 if(!in_array($u, $eventShareesNames)) {
                                                         $eventShareesNames[] = $u;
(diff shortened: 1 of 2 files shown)


georgehrke:
Code looks good, will test asap

@quamenzullo Thx for the pull request :)

lib/app.php
```diff
@@ -551,6 +551,18 @@ public static function sendEmails($eventid, $summary, $location, $description, $
 			}
 			foreach ($eventsharees as $sharee) {
 				$eventShareesNames[] = $sharee['share_with'];
+				$shwth = $sharee['share_with'];
+                                if(OC_Group::groupExists($shwth)) {
```
georgehrke: users and groups can have the same id, please check `$sharee['share_type']` instead

nicolashainaux:
I have too much work these days but I'll check soon what you've said here

lib/app.php
nicolashainaux: OK, I've checked it, corrected it on my server too, it works fine, and in addition I noticed I made an error in this commit (the line right above, 553, should not be there). I will create a new commit correcting this and pull a new request. This one can be removed.

nicolashainaux:
OK, I think I don't master the github way of managing pull requests, I added a new commit which corrects the errors. It seems it has been added to the pull request, though I am not completely sure. Tell me if I must do something else.

georgehrke:
Code looks good now, will test this asap. Thx for the pull request!

PR #561, opened 2014-09-15
Translatable remove calendar message and else styling

vgezer:
@georgehrke please review. thanks.

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -21,7 +21,7 @@ Calendar={
 			},
 			function(result){
 				if(result.status !== 'success'){
-					OC.dialogs.alert(result.data.message, 'Error sending mail');
+					OC.dialogs.alert(result.data.message, t('calendar', 'Error sending mail'));
 				}
 				Calendar.UI.loading(false);
 			}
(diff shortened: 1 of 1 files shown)


js/calendar.js
```diff
@@ -124,7 +124,8 @@ Calendar={
 		loading: function(isLoading){
 			if (isLoading){
 				$('#loading').show();
-			}else{
+			}
+			else {
```
georgehrke: please revert this change

georgehrke: as well as all the other } else to } \n else

vgezer: This makes the code search harder since most of the code uses this styling. Even in this file...?

vgezer: Just one example: https://github.com/owncloud/calendar/blob/minor/js/calendar.js#L128

georgehrke: http://doc.owncloud.org/server/6.0/developer_manual/app/general/codingguidelines.html#id2

vgezer: So do you think better i replace `}else{` with `} else {` for the same reason, or ignore them completely?

georgehrke: I'd say ignore them completely

georgehrke:
thx :)

PR #572, opened 2014-09-22
add documentation link

vgezer:
so that easier access in app management


georgehrke:
I feel like this pr should be made against stable7 as the documentation refers to ownCloud 7.

PR #588, opened 2014-10-13
add documentation link

vgezer:
Correct branch for https://github.com/owncloud/calendar/pull/572

PR #589, opened 2014-10-13
add documentation link

vgezer:
this is the master for https://github.com/owncloud/calendar/pull/588


georgehrke:
thx :smile:

LukasReschke:
It would be better to solve this using `go.php`: https://github.com/owncloud/documentation/blob/master/go.php

Then in the info.xml just link to the key. This works with master and stable7, will be released with 7.0.3.

PR #612, opened 2014-11-07
remote.php: authentication realm from config

semanino:
fixes #610 
optionally use core config item "instancename" for authentication realm instead of hard-coded string "ownCloud"

### DIFF
--- a/appinfo/remote.php
+++ b/appinfo/remote.php
@@ -41,7 +41,9 @@ $server = new \Sabre\DAV\Server($nodes);
 $server->httpRequest = $requestBackend;
 $server->setBaseUri($baseuri);
 // Add plugins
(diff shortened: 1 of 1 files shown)


georgehrke:
cc @LukasReschke

LukasReschke:
As stated in the issue: I'm not a fan of that and won't accept something like that into core... - What you do in Calendars/Contacts is your business ;-)

I just really don't see any advantage at all instead of the warm fluffly feeling being able to use an insecure SSL proxy.

georgehrke:
@semanino 
see @LukasReschke 's comment: https://github.com/owncloud/calendar/issues/610#issuecomment-62134304

You are free to apply this patch for your instance, but we are not willing to ship something that encourages those unsafe configurations.

PR #632, opened 2014-12-02
use OCP namespace, fix search

butonic:
fixes `Fatal error: Class 'OC\Search\Provider' not found in /.../core/apps-repos/calendar/lib/search/provider.php on line 26`


georgehrke:
thx

PR #639, opened 2014-12-06
Let users set a default calendar for new events [with hidden buttons]

vgezer:
:construction: 
Continuing from: https://github.com/owncloud/calendar/pull/399

I will need @jancborchardt's help at this stage since I am not a CSS-friendly guy :). Also if @Raydiation can help me this would be perfect.


js/calendar.js
```diff
@@ -628,6 +628,20 @@ Calendar={
 					  });
 				}
 			},
+			makeDefault:function(calid){
+
+			    $.post(OC.filePath('calendar', 'ajax/calendar', 'makedefault.php'), { calendarid: calid},
+				   function(data) {
+				     if (data.status == 'success'){
```
BernhardPosselt, line 635: #

templates/part.eventform.php
```diff
@@ -30,7 +30,7 @@
 			}
 		}
 		$sortedCalendars = array_merge($enabledCalendars, $disabledCalendars);
-		print_unescaped(OCP\html_select_options($sortedCalendars, $_['calendar'], array('value'=>'id', 'label'=>'displayname'))); ?>
+		print_unescaped(OCP\html_select_options($sortedCalendars, OCP\Config::getUserValue(OCP\USER::getUser(), 'calendar', 'defaultcalendar', $_['calendar']), array('value'=>'id', 'label'=>'displayname'))); ?>
```
BernhardPosselt, line 33: why print_unescaped? XSS

js/calendar.js
vgezer, line 635: The rest of your message is missing.

templates/part.eventform.php
vgezer, line 33: We might have some non-ASCII names in calendars. How do we display them?

js/calendar.js
BernhardPosselt, line 635: ah crap :)

I meant ===

templates/part.eventform.php
BernhardPosselt, line 33: Escaping will work fine for that. Escaping just means display it and escape HTML characters. I doubt that you want to use a username to create a div in the app, correct?

BernhardPosselt, line 33: PS: you should probably review the template code (and every print_unescaped). You should only use print_unescaped if you want to include html

vgezer, line 33: Ok, I will change this and other occurrences in the current file. However, for the rest of the code, I will submit another PR.

This was to select the default calendar in dropdown list.

PR #645, opened 2014-12-12
add file actions only when OCA.Files is available

butonic:
@PVince81 @georgehrke @raghunayyar a quick review pls

### DIFF
--- a/js/loader.js
+++ b/js/loader.js
@@ -176,8 +176,8 @@ Calendar_Import={
 	}
 }
 $(document).ready(function(){
-	if(typeof FileActions !== 'undefined'){
-		FileActions.register('text/calendar','importCalendar',  OC.PERMISSION_READ, '', Calendar_Import.Dialog.open);
-		FileActions.setDefault('text/calendar','importCalendar');
+	if(OCA.Files && OCA.Files.fileActions) {
+		OCA.Files.fileActions.register('text/calendar','importCalendar',  OC.PERMISSION_READ, '', Calendar_Import.Dialog.open);
+		OCA.Files.fileActions.setDefault('text/calendar','importCalendar');
 	};
 });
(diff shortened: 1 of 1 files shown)


PVince81:
Does this work from the "Shared with XXX" view ?
You might need to fix the open method of the handler to also read the "dir" property from the context. Because flat views like "Shared with XXX" need that extra info.
See here for an example: https://github.com/owncloud/music/blob/master/js/public/musicFilePlayer.js#L57

PVince81:
Note that the old `FileActions` only works for "All files" while the new `OCA.Files.FileActions` is designed to work with all file lists.

DeepDiver1975:
bildschirmfoto von 2015-10-12 14-04-20
- [ ] The action appears as 'Edit' without an icon
- [ ] Clicking Edit opens the text editor

georgehrke:
@DeepDiver1975 Is the OCA.Files.fileActions api documented somewhere?

PVince81:
not yet...

there is also another approach: https://github.com/owncloud/activity/blob/master/lib/fileshooksstatic.php#L43 and https://github.com/owncloud/activity/blob/master/lib/fileshooksstatic.php#L98

This would only load the JS files when the file apps is visible.

PVince81:
(only works in 8.2)

DeepDiver1975:
I'm fixing this for 8.2

georgehrke:
Thanks :blush:

DeepDiver1975:
Ah - the Edit is from the text editor ..... will submit PR there

DeepDiver1975:
> Ah - the Edit is from the text editor ..... will submit PR there

@jancborchardt @PVince81 I assume it is intended to have an icon on the action menu

DeepDiver1975:
@PVince81 please review - THX

PVince81:
Looks good :+1:

Issue #661, opened 2015-01-03
Share icon not available in Firefox


codecrusher64:
After an update of Firebox browser, the share icon is not available anymore. So I can not change the sharing settings now.

![owncloud_calender_share](https://cloud.githubusercontent.com/assets/7566180/5603753/db57931c-9394-11e4-994d-4bb26a91f3c8.png)

**Platform information:**
* _OS:_ Windows 8.1
* _Owncloud:_ 7.0.2
* _Firefox:_ 34.0.5

georgehrke:
@jbtbnl Do you have some spare time to maybe look into this?

RealRancor:
Hi,

IRC chatlog of another user from yesterday which noticed the same:

```
Guest31704: [02:42:59] yes, that's where I'd expect to see it. On the far left of the icon set. But instead of 5 icons:  share / globe / download / edit / delete I just have 4, no share icon, and instead of a globe I have a "link" as the icon furthest left. the other 3 icons match the documentation.
Guest31704: [02:46:32] ok... I tried looking at it from another web browser... IE instead of firefox... and the share icon is present in IE.
Guest31704: [02:49:30] ok... seems that the share icon was being blocked by adblock plus. clearly not a problem with owncloud, and I'll have to dig deeper to find out what triggered adblock to remove it. cheers.
```

Aschen:
Hi,

I experienced the same issue with Owncloud 7.0.3 and Firefox 34.0.5 on Debian 7.

georgehrke:
please help testing https://github.com/owncloud/calendar/pull/662, thx

RealRancor:
Reasons for this was posted a few minutes ago at https://github.com/owncloud/core/issues/8360#issuecomment-69195136

Seems a default-enabled ABP-list „Fanboy's Social Blocking List“ and all the other parts of OC are also affected.

PR #662, opened 2015-01-06
Fix share icon for Firefox

jbtbnl:
@codecrusher64 @georgehrke @RealRancor please check out this fix for #661.

I tested it on Chrome and Firefox on Ubuntu, please contribute your findings on other platforms.


codecrusher64:
I checked it on my platform with Firefox and Chrome. Both show share icon correctly now.

Aschen:
Ok with Firefox 34.0.5 on Windows 7.

Thanks

jbtbnl:
@georgehrke should this be backported?

georgehrke:
yes, I'll take care

jancborchardt:
Wait – icon-shared is a different icon than icon-share! And one we shouldn’t use anymore.

jancborchardt:
We need to fix this in AdBlock and let the Adblock devs know to exclude that.

ghost:
Yes, as core is also affected by this Adblock rule:

https://github.com/owncloud/core/issues/8360#issuecomment-69195136

PR #671, opened 2015-01-14
fix js and css loading for owncloud 8

BernhardPosselt:
Moves all scripts and styles to index.php so that they are only loaded when the calendar is actually opened and not when you launch the News app for instance. Also the first paramter to addScript is the application name and this breaks various things in core if you just use a hack like the path

@georgehrke @raghunayyar

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -40,7 +40,7 @@ Sabre\VObject\Property::$classMap['LOCATION'] = 'OC\VObject\StringProperty';
 
 $url = OC::$server->getRequest()->server['REQUEST_URI'];
 
-if (preg_match('%index.php/apps/files/?.*%', $url)) {
+if (preg_match('%index.php/apps/files(/.*)?%', $url)) {
     OCP\Util::addscript('calendar','loader');
     OCP\Util::addScript('calendar', '../3rdparty/chosen/js/chosen.jquery.min');
     OCP\Util::addStyle('calendar', '../3rdparty/chosen/css/chosen');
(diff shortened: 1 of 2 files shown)


georgehrke:
everything in appinfo/app.php is needed for the ical import in the files app. So moving them to index.php will break the import dialog.

PVince81:
Isn't that a case where addVendorScript() can be used ?
CC @MorrisJobke

georgehrke:
I agree that we should load the chosen and minicolor js/css after the ical file was clicked, so we only load the calendar/js/loader.js file

BernhardPosselt:
@georgehrke separate template? If so it should be added there

georgehrke:
add it where? in the files app? That's not a real solution as all apps (including 3rdparty ones) are supposed to register file handers for the files app

BernhardPosselt:
Ah, files app. Why do we support this again?

BernhardPosselt:
We should a wrapper then like if ($currentApp === 'files')

georgehrke:
> We should a wrapper then like if ($currentApp === 'files')

That would be a good work-around for now. Could you add it to your pr please?

BernhardPosselt:
Done I simply preg_match the url since there is no working api that gets you the currently active app. If the files app was written using the app framework we could match it using the route id

georgehrke:
awesome thanks!, will test asap

DeepDiver1975:
tested works :+1:

PVince81:
:+1: works

Issue #530, opened 2014-08-17
v7 calender throws JavaScript Error


SunboX:
```javascript
TypeError: $(...).offset(...) is undefined calendar.js:120
TypeError: $(...).offset(...) is undefined calendar.js:120
```
Line 120:
```javascript
timeline.css({'left':($('.fc-today').offset().left-appNavigationWidth),'width': $('.fc-today').width(),'top':topLoc + 'px'});
```

carstengumz:
Same issue (calendar not loading, error "$(...).offset(...) is undefined" in calendar.js, line 120) here after upgrade from v6.0.4 to v7.0.4. Not all users where affected. So I looked for differences in their settings. 

Finally, I was able to fix it by performing this database (MySQL) update:

```sql
UPDATE owncloud.oc_preferences 
SET configvalue = 'agendaDay'
WHERE appid = 'calendar' AND
      configkey = 'currentview' AND
      configvalue = 'list';
```
I guess this is something the updater should have done during the update process.

georgehrke:
@carstengumz Thanks a lot for the hint with the list view!

georgehrke:
please help testing: https://github.com/owncloud/calendar/pull/675

basc:
I have executed @carstengumz SQL Script, but the error still occurs.

See me post from 17.Dez.2014
```
Error occurs every time the function setTimeline() is called (every 60 seconds)
on every view (day, week, month) but only if "today" is not present.
i.e. goto next week or month and after 60 seconds the error is in the javascript console.
```

Does nobody get this?
I have this error after 60 seconds in javascript console if "today" is not visible.

basc:
Another way to reproduce the error.

The function setTimeline() is called every 60 seconds.
If you don't want to wait then execute the following line in javascript console:
Calendar.Util.setTimeline();

But the error only occurs if you first navigate away from today, which means i.e. in month view go forward or backward one month. now is january, go forward to february or back to dezember,
only there the error occurs. the same in week or day view.

The error:
Line 120 in calendar.js
```
timeline.css({'left':($('.fc-today').offset().left-appNavigationWidth),'width': $('.fc-today').width(),'top':topLoc + 'px'});
```
$('.fc-today') is undefined because the table-cell with class "fc-today" is not present if you don't watch the current month/week/day.

Can you now reproduce the error?

brantje:
Yea thank, i have a find in mind. Currently at work, will try this weekend.

[...]

PR #675, opened 2015-01-18
list ain't a valid value for currentview anymore

georgehrke:
fixes https://github.com/owncloud/calendar/issues/530


basc:
#675 doesn't fix #530 
See my post there

PR #688, opened 2015-01-27
Remove isShipped attribute

nickvergessen:
@DeepDiver1975 

owncloud/core#13706

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -5,9 +5,7 @@
 	<licence>AGPL</licence>
 	<author>Georg Ehrke, Bart Visscher, Jakob Sack</author>
 	<require>8</require>
-	<shipped>true</shipped>
 	<description>Calendar with CalDAV support</description>
-	<default_enable/>
 	<remote>
 		<calendar>appinfo/remote.php</calendar>
 		<caldav>appinfo/remote.php</caldav>


appinfo/info.xml
```diff
@@ -5,7 +5,6 @@
 	<licence>AGPL</licence>
 	<author>Georg Ehrke, Bart Visscher, Jakob Sack</author>
 	<require>8</require>
-	<shipped>true</shipped>
 	<description>Calendar with CalDAV support</description>
 	<default_enable/>
```
georgehrke: please remove this line as well, this will cause issues when installing the app over the app store

nickvergessen: I assume this applies to all apps?

georgehrke: Ok, so there is no check inside [checkAppsIntegrity](https://github.com/owncloud/core/blob/9771ead7dd0f761e5379739d9bf53048ba602fe8/lib/private/installer.php#L286), but it's better to remove it anyway

nickvergessen: right, removed it

georgehrke: Dev doc says that it's for core apps only http://doc.owncloud.org/server/6.0/developer_manual/app/appframework/info.html#default-enable

georgehrke:
:+1:

Issue #554, opened 2014-09-11
Time looks ungly on mac


brantje:
![screen shot 2014-09-11 at 10 56 33](https://cloud.githubusercontent.com/assets/2758767/4231486/a8661990-3991-11e4-8aee-a83f09ee7b48.png)
Can we change the input type from time to text?
Apple messes up the display.

brantje:
I've only got an Apple device at work, and can't do ownCloud this at work. 
Sorry. But we can test with some test tools?

raghunayyar:
Self assigning. I have a mac.

PR #717, opened 2015-02-14
Fix: Adds widths to time inputs in event dialog

raghunayyar:
A fix for https://github.com/owncloud/calendar/issues/554

@owncloud/designers @georgehrke


jancborchardt:
Nice one!

PR #653, opened 2014-12-19
Fix namespace of Sabre classes

nickvergessen:
- [ ] Depends on owncloud/core#12864

@DeepDiver1975 @georgehrke

PR #722, opened 2015-02-16
Update sabre to 2.1

PVince81:
Depends on core's sabre update: https://github.com/owncloud/core/pull/14151

This at least lets the calendar app load.
I haven't tested whether CalDAV still works.

Also, do you have an idea what "Principals" are ?

@georgehrke

### DIFF
--- a/lib/sabre/backend.php
+++ b/lib/sabre/backend.php
@@ -1,5 +1,7 @@
 <?php
 
+use \Sabre\DAV\PropPatch;
+
 class OC_Connector_Sabre_CalDAV extends \Sabre\CalDAV\Backend\AbstractBackend {
 	/**
 	 * List of CalDAV properties, and how they map to database fieldnames
@@ -172,10 +174,10 @@ class OC_Connector_Sabre_CalDAV extends \Sabre\CalDAV\Backend\AbstractBackend {
 	 * (424 Failed Dependency) because the request needs to be atomic.
 	 *
 	 * @param string $calendarId
-	 * @param array $properties
+	 * @param PropPatch $propPatch
 	 * @return bool|array
 	 */
-	public function updateCalendar($calendarId, array $properties) {
+	public function updateCalendar($calendarId, PropPatch $propPatch) {
 
 		$newValues = array();
 		$result = array(
@@ -185,6 +187,8 @@ class OC_Connector_Sabre_CalDAV extends \Sabre\CalDAV\Backend\AbstractBackend {
 		);
 
 		$hasError = false;
+
+		$properties = $propPatch->getRemainingMutations();
 
 		foreach($properties as $propertyName=>$propertyValue) {
 
@@ -409,7 +413,7 @@ class OC_Connector_Sabre_CalDAV extends \Sabre\CalDAV\Backend\AbstractBackend {
 	/**
 	 * @brief Creates a etag
 	 * @param array $row Database result
-	 * @returns associative array
+	 * @returns array associative array
 	 *
 	 * Adds a key "etag" to the row
 	 */
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -34,10 +34,6 @@ OCP\App::addNavigationEntry( array(
 OCP\Share::registerBackend('calendar', 'OC_Share_Backend_Calendar');
 OCP\Share::registerBackend('event', 'OC_Share_Backend_Event');
 
-Sabre\VObject\Property::$classMap['SUMMARY'] = 'OC\VObject\StringProperty';
-Sabre\VObject\Property::$classMap['DESCRIPTION'] = 'OC\VObject\StringProperty';
-Sabre\VObject\Property::$classMap['LOCATION'] = 'OC\VObject\StringProperty';
-
 $request = \OC::$server->getRequest();
 if (isset($request->server['REQUEST_URI'])) {
 	$url = $request->server['REQUEST_URI'];
--- a/appinfo/remote.php
+++ b/appinfo/remote.php
@@ -19,13 +19,12 @@ if(\OCP\App::isEnabled('contacts')) {
 }
 
 // Backends
-$authBackend = new OC_Connector_Sabre_Auth();
+$authBackend = new \OC\Connector\Sabre\Auth();
 $principalBackend = new \OC\Connector\Sabre\Principal(
 	\OC::$server->getConfig(),
 	\OC::$server->getUserManager()
 );
 $caldavBackend    = new OC_Connector_Sabre_CalDAV();
-$requestBackend = new OC_Connector_Sabre_Request();
 
 // Root nodes
 $Sabre_CalDAV_Principal_Collection = new \Sabre\CalDAV\Principal\Collection($principalBackend);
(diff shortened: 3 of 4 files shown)


PVince81:
Also I'm not 100% sure whether those lines must be removed or replaced with others.
I thought that the new Sabre was automatically mapping these to internal string properties that have all the necessary fixes.

Note that StringProperty was removed in core, see https://github.com/owncloud/core/pull/14151#discussion-diff-24604921 and https://github.com/owncloud/core/pull/14151#discussion-diff-24604937

georgehrke:
StringProperty was introduced by tanghus. IIRC it was to fix escaping issues that incorrectly replaced , to \, etc. 
I am not sure if this was fixed in Sabre VObject, but I guess so.

PVince81:
Might help: http://sabre.io/vobject/upgrade/
But not sure what the `$classMap` replacement is.

PVince81:
@evert I don't see any mention of `Sabre\VObject\Property::$classMap` in the update guide here: http://sabre.io/vobject/upgrade/

Or was `$classMap` not the correct way in the first place ?

georgehrke:
If I am not completely mistaken `$classmap` has been split up into `$propertyMap` and `$componentMap`

georgehrke:
but nevertheless: If the escaping issue (, to \,) was fixed in upstream (which I am pretty sure about), there is no need for classmap respectively propertyMap in the calendar app anymore.

PVince81:
Ok cool. I'll still need to figure it out for the contacts app though.

Can you help testing if CalDAV still works ?

georgehrke:
> Can you help testing if CalDAV still works ?

Sure, will try to find some time in the next couple of days.

PVince81:
Take your time. The core PR still needs some work :smile:

evert:
@PVince81 , the classMap indeed went to two new properties, and they can be found in `Sabre\VObject\Component\VCalendar` and `Sabre\VObject\Component\VCard`, but as @georgehrke said, I believe you no longer need them... and if you do, it's probably wiser to try to work around it.

evert:
Principals are users and groups in webdav terms. Any caldav or carddav user must also have a principal node that the client uses to fetch information about the user, as well as a few other things.

PVince81:
I had a closer look and it looks like updating the calendar app will be more than just tweaking stuff.

A few things that will need to be adapted:
- [x] replace "Component::create('VCALENDAR')" with "new VCalendar()"
- [x] replace "Component::create('VEVENT') with "$vcard->createComponent('VEVENT')"
- [ ] fix date time handling everywhere (getDateType() is gone, getDateTimeZone() seems to be different too)
- [x] update Sabre backends to work with the new PropFind/PropPatch objecst
- [x] update name of sabre classes that came from core (OC_Connector_Sabre_XXX => OC\Connector\Sabre\XXX)

georgehrke:
will take care of the changes

PVince81:
I have a few changes locally (VCard), will push soon.
I started with the DateTime but I got confused and is messy, so I'll revert.

georgehrke:
> I have a few changes locally (VCard), will push soon.

Ok, but VCard sounds wrong. Is there no VCalendar class?

PVince81:
@georgehrke ideally we should try avoid using OC_VObject from core altogether, if possible.

georgehrke:
Yes, I totally agree, but VCards are for contact-cards. There is probably a dedicated VCalendar class (in vobject)

PVince81:
Ah sorry, I mixed them up. It's VCalendar that needs instantiating.

I pushed my change, feel free to take over now.

georgehrke:
> I pushed my change, feel free to take over now.

Thanks, will do

osos:
Will this automitically fix calendar invites, issue #318

PVince81:
@georgehrke any update ? Do you need help ?

I was able to fix the contacts app so I might be able to help a bit. My only problem here is that I don't understand all these different date formats (and why they are there) and how to convert them to the new API.

georgehrke:
Had a lot of things to do for university last 2 weeks, should find some time this week

PVince81:
Ok cool. Please shout if you need help.

georgehrke:
Note to myself:
- `getDateType` the way we use it can be replaced with `isFloating`
- `getDateTimeZone` with $element['TZID']

lib/sabre/backend.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use \Sabre\DAV\PropPatch;
+
 class OC_Connector_Sabre_CalDAV extends \Sabre\CalDAV\Backend\AbstractBackend {
```
nickvergessen, line 5: Please fix the namespace while on it.
See https://github.com/owncloud/calendar/pull/653 (change namespace and remove classpath hack in app.php)

DeepDiver1975:
@georgehrke please try to finish this asap - we need to provide an update for 8.1 on the appstore.
Still 4 weeks to go - but fixing this asap does not hurt :wink:

PVince81:
Rebased

PVince81:
@georgehrke how about merging this first ?
At least with this PR it is possible to enable the app (no fatal error).

Then continue in separate PRs for the date management and other things ?

PVince81:
I rechecked the TODO list from https://github.com/owncloud/calendar/pull/722#issuecomment-75223051 and it looks like I was already done with VCalendar and VEVENT. So the only remaining stuff is converting the dates (and of course extensive testing/bugfixing)

nickvergessen:
@PVince81 mint fixing https://github.com/owncloud/calendar/pull/722#discussion_r27372630 before merging? ;)

PVince81:
This can be merged afterwards.

PVince81:
Ok, seems I was confused by some methods missing in the API. They're not missing but were custom methods in core's "VObject" override. Let's see how to replace these.

PVince81:
@evert is this:

```
$vevent->DTSTART = $now;
```

the same as

```
$vevent->create('DTSTART')->setValue($now);
```

?

What happens if I call create twice on the same property, will it be duplicated ?

PVince81:
There are too many ways to do the same thing, this is very confusing.

Trying to figure out how to properly set VEVENT properties. I had $vevent->DTSTART->setValue() is not defined, maybe I need to create the property first.

PVince81:
Looks like the $vevent object didn't have the properties by default, this did the trick:

``` php
            $start = new DateTime($from.' '.$fromtime, $timezone);
            $end = new DateTime($to.' '.$totime, $timezone);
            $start->setTimeZone($timezone);
            $end->setTimeZone($timezone);
            $dtstart = $vcalendar->create('DTSTART');
            $vevent->DTSTART = $dtstart;
            $dtend = $vcalendar->create('DTEND');
            $vevent->DTEND = $dtend;
            $dtstart->setValue($start);
            $dtend->setValue($end);
```

georgehrke:
Yes, we should merge this, so the fatal error is gone and you can at least use your ownCloud instance when the calendar is installed.

PVince81:
Ok, will merge this first. I have some local changes for the dates, will raise separately.

PVince81:
Continuing here: https://github.com/owncloud/calendar/pull/771

evert:
Hey @PVince81, your last code snippet:

```
        $start = new DateTime($from.' '.$fromtime, $timezone);
        $end = new DateTime($to.' '.$totime, $timezone);
        $start->setTimeZone($timezone);
        $end->setTimeZone($timezone);
        $dtstart = $vcalendar->create('DTSTART');
        $vevent->DTSTART = $dtstart;
        $dtend = $vcalendar->create('DTEND');
        $vevent->DTEND = $dtend;
        $dtstart->setValue($start);
        $dtend->setValue($end);
```

Is rather dense =). This _should_ also work:

```
        $start = new DateTime($from.' '.$fromtime, $timezone);
        $end = new DateTime($to.' '.$totime, $timezone);

        $vevent->DTSTART = $start;
        $vevent->DTEND = $end;
```

If that's too magic:

```
        $start = new DateTime($from.' '.$fromtime, $timezone);
        $end = new DateTime($to.' '.$totime, $timezone);

        $vevent->add('DTSTART', $start);
        $vevent->add('DTEND', $end);
```

The difference between my first and second replacement is that the first _sets_ the DTSTART, and the latter _adds_ a property. So the latter would add a secondary `DTSTART` property if one already existed.

georgehrke:
Applied @evert's suggestions in https://github.com/owncloud/calendar/commit/aa71f8f4b29e4c913eb58846379cecc9b50533cc

PR #737, opened 2015-02-24
fix event dialog position and size on mobile, fix #703

### LINKED ISSUE #703: UI for calendar on a mobile device has event dialog box under everything else
Hi,

I have a very hard time editing/adding an event from a mobile device as the CSS makes the dialog box disappear under the calendar, but not the sidebar. I think the z-index is totally wrong. I consider this a serious issue. I also get this in other browsers, not just Firefox.

jancborchardt:
@ericcotelnu this should fix https://github.com/owncloud/calendar/issues/703

Please also review @owncloud/designers @georgehrke @brantje 

We should also backport this cause it’s a pretty serious bug which prevents from creating events on mobile.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -251,6 +251,8 @@ button.category{margin:0 3px;}
 	overflow-y: auto;
 	overflow-x: hidden;
 	padding: 0;
+	z-index: 1000;
+	max-width: 100%;
 }
 .ui-dialog .ui-dialog-content {
 	overflow: initial;


georgehrke:
code changes look good, will test tomorrow

MorrisJobke:
:+1:

jancborchardt:
@georgehrke backport necessary here as well. :)

Issue #714, opened 2015-02-11
 Undefined index: REQUEST_URI


tony5:
I am running debian 7
everything was working fine with OC 7.0.4
after upgrading to OC 8.0 I get an error in my logs that is triggered everytime cron runs.
the error is:
 Undefined index: REQUEST_URI at /var/www/owncloud/apps/calendar/appinfo/app.php#41 
I never had this error on OC 7.0.4
I have tried to disable/remove/reinstall but I still get the same error.
other people have done a fresh install and get the error as well.

any advice?
take care

georgehrke:
please use the [issue template](https://raw.github.com/owncloud/core/master/issue_template.md) as described in our [contributing guidelines](https://github.com/owncloud/core/blob/master/CONTRIBUTING.md)

Panoptis:
Same here
Error PHP Undefined index: REQUEST_URI at /var/www/virtual/.../apps/calendar/appinfo/app.php#41
Calendar app seems to work fine though.
OC 8.0 (stable)
PHP 5.6.5 on CentOS

georgehrke:
@Panoptis please use the issue template. I need more information for this!

georgehrke:
>  cron runs.

What does the cron call look like?

tony5:
### Steps to reproduce
1. upgrade OC from 7.0.4 to OC 8.0
2. Reinstall the app
3. do a fresh install

### Expected behaviour
Tell us what should happen
the error  Undefined index: REQUEST_URI at /var/www/owncloud/apps/calendar/appinfo/app.php#41 should not be in the logs

### Actual behaviour
 Undefined index: REQUEST_URI at /var/www/owncloud/apps/calendar/appinfo/app.php#41 is in the logs and run by system cron every 15 minutes.

### Server configuration
debian 7

**Web server:**
apache2
**Database:**
mysql
**PHP version:**
PHP Version 5.4.36-0+deb7u3
**ownCloud version:** (see ownCloud admin page)
 ownCloud 8.0 (stable) 
**Updated from an older ownCloud or fresh install:**
yes 7.0.4
**List of activated apps:**
Activity 1.2.0, Calendar 0.6.4,
Contacts 0.3.0.18, 
Deleted files 0.6.2, 
Documents 0.9.0, 
External storage support 0.2.3,
Mail Template Editor 0.1
First Run Wizard 1.1
File Locking
News 5.2.2
PDF Viewer 0.7
Pictures 0.6.0
[...]

PR #752, opened 2015-03-18
Fix error message for cron run

MorrisJobke:
For every cron run there is no request URI set. This checks for the existence before accessing it.

backport to stable8 would be nice, because this pollutes the log.

cc @georgehrke @Raydiation 

To test:
- Activate calendar
- run php -f cron.php before and after this

Error message was:

```
{"reqId":"3b72eb1fd612fa6bc4f01721ca6dddfa","remoteAddr":"","app":"PHP","message":"Undefined index: REQUEST_URI at \/home\/mjob\/Projekte\/owncloud\/master\/apps2\/calendar\/appinfo\/app.php#41","level":0,"time":"2015-03-18T09:50:56+00:00","method":"--","url":"--"}
```

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -38,12 +38,15 @@ Sabre\VObject\Property::$classMap['SUMMARY'] = 'OC\VObject\StringProperty';
 Sabre\VObject\Property::$classMap['DESCRIPTION'] = 'OC\VObject\StringProperty';
 Sabre\VObject\Property::$classMap['LOCATION'] = 'OC\VObject\StringProperty';
 
-$url = OC::$server->getRequest()->server['REQUEST_URI'];
+$request = \OC::$server->getRequest();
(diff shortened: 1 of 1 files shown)


MorrisJobke:
Whitespace cleaned diff helps a lot: https://github.com/owncloud/calendar/pull/752/files?w=1

georgehrke:
I already provided a patch in https://github.com/owncloud/calendar/issues/714 . Just need to add to master and stable8.

Please excuse my brevity and typos.

MorrisJobke:
@georgehrke Please don't do it in the dirty way with adding a `@` in front of the statement. :(

georgehrke:
> @georgehrke Please don't do it in the dirty way with adding a `@` in front of the statement. :(

True, your approach is better than my laziness. Let's use yours

georgehrke:
@MorrisJobke thx!

PR #758, opened 2015-03-26
Don't override default today highlight colour with JS

### LINKED ISSUE #757: Today highlight colour - js
It seems this issue has already been raised in #45 and the CSS was changed accordingly.

I am still having a problem viewing the today calendar though, since the Javascript is still setting it as #ffa
See here: https://github.com/owncloud/calendar/blob/master/js/calendar.js#L1046

That line should be changed to #ffa as well

pjrobertson:
Fixes #757 

I decided in the end that it would be wiser to remove any custom-set attribute all together, and fall back to the one defined [in style.css](https://github.com/owncloud/calendar/blob/master/css/style.css#L113)

CC @georgehrke

### DIFF
--- a/js/calendar.js
+++ b/js/calendar.js
@@ -1043,7 +1043,8 @@ $(document).ready(function(){
 			if(view.name === 'agendaDay') {
 				$('td.fc-state-highlight').css('background-color', '#ffffff');
 			} else{
-				$('td.fc-state-highlight').css('background-color', '#ffc');
+				// remove any custom CSS. Fall back to default for .fc-state-highlight (style.css)
+				$('td.fc-state-highlight').css('background-color', '');
 			}
 			Calendar.UI.setViewActive(view.name);
 			if (view.name == 'agendaWeek') {


georgehrke:
looks good to me :+1: 

please help reviewing @jancborchardt

jancborchardt:
Calendar is broken for me currently for core master so I can’t test. :\

@pjrobertson can you post a before-after screenshot?

pjrobertson:
Hmmm... there's really not much to see, and it's all standard CSS, so I'm not quite sure that's necessary!

If you just change the JS on any install of owncloud you have (even just change it in your browser) you'll see that all this does is remove the `style="background-color:#ffc"` from the 'today' box

georgehrke:
@jancborchardt you can also apply the change to stable8

georgehrke:
ping @jancborchardt

jancborchardt:
I’m sitting in the sun and the current day highlight color is pretty fine. The color in this pull request makes it much too yellow and a tad obnoxious. :\

sixtyfive:
So the "today" highlight color in the calendar is still too light, which might have been forgotten about, since the CSS was apparently changed and the issue (#45) was closed. Now there's another open issue (#757) because of the JS override and then there's this pull request here where nothing seems to be happening because of confused feelings about the effing colour. The original change was from #FFC to #FFA. Couldn't you just change it to that for now?

DeepDiver1975:
@jancborchardt please have a look - change seems valid to me

jancborchardt:
Strange, current master gives me this, which seems to be what is achieved by this PR, no?
capture du 2015-10-13 09-31-20

So @pjrobertson or @sixtyfive if it still occurs to you, could you submit a pull request with the color change from #ffc to #ffa? Because that’s way better than the color in this pull request.

DeepDiver1975:
> Because that’s way better than the color in this pull request.

and ... closing

PR #763, opened 2015-04-08
Prevent remote.php to be used in maintenance mode

PVince81:
@georgehrke

Note: not testable yet due to the sabre upgrade

### DIFF
--- a/appinfo/remote.php
+++ b/appinfo/remote.php
@@ -50,6 +50,7 @@ $server->addPlugin(
 $server->addPlugin(
     new Sabre\CalDAV\Schedule\IMipPlugin($adminmail)
     );
+$server->addPlugin(new \OC\Connector\Sabre\MaintenancePlugin());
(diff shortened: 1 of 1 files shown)


appinfo/remote.php
```diff
@@ -44,6 +44,7 @@
 $server->httpRequest = $requestBackend;
 $server->setBaseUri($baseuri);
 // Add plugins
+$server->addPlugin(new \OC\Connector\Sabre\MaintenancePlugin());
```
nickvergessen, line 53: Should we create a version of `\Sabre\DAV\Server` in core which automatically adds plugins like this?

PVince81, line 53: something to look into for 8.2.

Maybe core must only provide a utility method / service class where this app calls "getServerInstance()". Then core would pre-initialize it.
Note that the instance for public.php might be different, so might need an additional "getPublicServerInstance()"

PR #771, opened 2015-04-16
Update sabre2.1 dates

PVince81:
The show goes on...

### DIFF
--- a/ajax/event/edit.form.php
+++ b/ajax/event/edit.form.php
@@ -21,45 +21,47 @@ if(!$data) {
 	OCP\JSON::error(array('data' => array('message' => OC_Calendar_App::$l10n->t('Wrong calendar'))));
 	exit;
 }
-$object = OC_VObject::parse($data['calendardata']);
+$object = \Sabre\VObject\Reader::read($data['calendardata']);
 $vevent = $object->VEVENT;
 $object = OC_Calendar_Object::cleanByAccessClass($id, $object);
-$accessclass = $vevent->getAsString('CLASS');
+$accessclass = $vevent->CLASS;
 $permissions = OC_Calendar_App::getPermissions($id, OC_Calendar_App::EVENT, $accessclass);
 
 $dtstart = $vevent->DTSTART;
 $dtend = OC_Calendar_Object::getDTEndFromVEvent($vevent);
-switch($dtstart->getDateType()) {
-	case Sabre\VObject\Property\DateTime::UTC:
+
+// DATE
+if ($dtstart->hasTime()) {
+	// UTC ?
+	if (!$dtstart->isFloating()) {
 		$timezone = new DateTimeZone(OC_Calendar_App::getTimezone());
-		$newDT    = $dtstart->getDateTime();
+		$newDT = $dtstart->getDateTime();
 		$newDT->setTimezone($timezone);
 		$dtstart->setDateTime($newDT);
-		$newDT    = $dtend->getDateTime();
+		$newDT = $dtend->getDateTime();
 		$newDT->setTimezone($timezone);
 		$dtend->setDateTime($newDT);
-	case Sabre\VObject\Property\DateTime::LOCALTZ:
-	case Sabre\VObject\Property\DateTime::LOCAL:
-		$startdate = $dtstart->getDateTime()->format('d-m-Y');
-		$starttime = $dtstart->getDateTime()->format('H:i');
-		$enddate = $dtend->getDateTime()->format('d-m-Y');
-		$endtime = $dtend->getDateTime()->format('H:i');
-		$allday = false;
-		break;
-	case Sabre\VObject\Property\DateTime::DATE:
-		$startdate = $dtstart->getDateTime()->format('d-m-Y');
-		$starttime = '';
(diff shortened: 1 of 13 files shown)


PVince81:
To test:
- [x] retest access CLASS
- [x] all day events
- [x] floating events
- [x] events created with different timezones

Event actions:
- [x] create event
- [x] delete event
- [x] move event
- [x] edit event
- [x] resize event (? how to trigger this ?)
- [x] import

georgehrke:
Thx a lot for helping out here. :)

PVince81:
- [x] double check calls to `$dt->modify()` as they don't seem to modify the original object any more

georgehrke:
>  resize event (? how to trigger this ?)

Go to the border of an event, an arrow is supposed to appear.
It's broken, I'm on it. (resizing is broken, not the arrow)

georgehrke:
import broken as well, on it

PVince81:
I fixed it, wait

georgehrke:
ok, thx

PVince81:
Calendar and event import work now.

For the timezone events I'm not sure, I see this happening:
1. Switch timezone to Shanghai
2. Create event between 09:00-11:00
3. Switch timezone to Berlin
4. The calendar view still puts the event at 09:00-11:00 but the edit view shows the correct values

Can you have a look ? I don't know if it was working like this before.

I'm done for now, feel free to take over.

georgehrke:
> Can you have a look ? I don't know if it was working like this before.

Will do. Thx a lot for your help :)

georgehrke:
The timezone stuff was a caching issue ...

ajax/event/edit.form.php
```diff
@@ -21,45 +21,47 @@
...
+	$enddate = $dtend->getDateTime()->format('d-m-Y');
+	$endtime = '';
+	$allday = true;
 }
 
-$summary = strtr($vevent->getAsString('SUMMARY'), array('\,' => ',', '\;' => ';'));
-$location = strtr($vevent->getAsString('LOCATION'), array('\,' => ',', '\;' => ';'));
-$categories = $vevent->getAsString('CATEGORIES');
-$description = strtr($vevent->getAsString('DESCRIPTION'), array('\,' => ',', '\;' => ';'));
+$summary = strtr($vevent->SUMMARY, array('\,' => ',', '\;' => ';'));
```
evert, line 61: `strtr` is _not_ needed. This is all done by sabre/vobject so you would double-encode it.

evert, line 61: Actually.. you are doing the opposite.. you are decoding the value. But yea this is also not needed _unless_ the original is incorrectly encoded. I would avoid it if you can.

lib/object.php
```diff
@@ -124,11 +124,12 @@ public static function add($id,$data) {
 				);
 			}
 		}
-		$object = OC_VObject::parse($data);
+		$object = Sabre\VObject\Reader::read($data);
 		list($type,$startdate,$enddate,$summary,$repeating,$uid) = self::extractData($object);
 
 		if(is_null($uid)) {
-			$object->setUID();
+			$uid = substr(md5(rand().time()), 0, 10);
```
evert, line 131: You can also use `Sabre\VObject\UUIDUtil::getUUID();`

PVince81:
@evert thanks for the comments. Maybe there are a few legacy stuff that needs to be cleant up.

@georgehrke can you take care of these extra fixes or do you want me to do them ? (I'd rather you do as you have better understanding of the original code)

Thanks.

lib/object.php
```diff
@@ -1055,38 +1057,56 @@ public static function updateVCalendarFromRequest($request, $vcalendar) {
...
 			$repeat = "false";
 		}
 
+		$now = new DateTime('now');
+		$now->setTimeZone(new \DateTimeZone('UTC'));
+		$lastModified = $vevent->__get('LAST-MODIFIED');
+		if (is_null($lastModified)) {
+			$lastModified = $vevent->add('LAST-MODIFIED');
+		}
+		$lastModified->setValue($now);
```
PVince81, line 1072: @evert is there a better way to do this ?
Because "$vevent->LAST-MODIFIED" is a syntax error, so I need to use "$vevent->__get('LAST-MODIFIED')" which might not exist.
It looks like direct access/assignment usually auto-creates the property ?

PVince81, line 1072: I don't want to add/create a new one if there is already one as it seems it's possible to duplicate properties.

evert, line 1072: The PHP syntax to access properties that contain special characters is :

```
$vevent->{'LAST_MODIFIED'} = $now;
```

Alternatively:

```
$vevent->__set('LAST_MODIFIED', $now);
```

evert, line 1072: oops those should be dashes, not underscores.

PVince81, line 1072: Cool, thanks. I didn't know about this construct :smile:

evert, line 1072: me neither until I already built that API and realized some properties have special characters in them ;) should probably document that better. I was hoping that a lot of the vobject api's would be 'natural' to people, but realize here again that that might not be true =(

Issue #769, opened 2015-04-16
[stable8] Missing or invalid fields


PVince81:
1. Setup v8.0.2 or stable8
2. Enable calendar app from the stable8 branch
3. Create an event:

![calendar_missingfields](https://cloud.githubusercontent.com/assets/277525/7182328/e6e231da-e450-11e4-94c4-6c7dec7b7186.png)

Log says this:
```
{"reqId":"67b32ffd5c5204b623b4c96de9714075","remoteAddr":"127.0.0.1","app":"PHP","message":"Undefined index: calendar at \/srv\/www\/htdocs\/owncloud\/apps3\/calendar\/ajax\/event\/new.php#21","level":3,"time":"2015-04-16T13:54:58+00:00","method":"POST","url":"\/owncloud\/index.php\/apps\/calendar\/ajax\/event\/new.php"}
{"reqId":"67b32ffd5c5204b623b4c96de9714075","remoteAddr":"127.0.0.1","app":"PHP","message":"Undefined index: userid at \/srv\/www\/htdocs\/owncloud\/apps3\/calendar\/lib\/object.php#117","level":3,"time":"2015-04-16T13:54:58+00:00","method":"POST","url":"\/owncloud\/index.php\/apps\/calendar\/ajax\/event\/new.php"}
```

Note: I have OC 8.0.3RC1 with calendar 0.6.4 and it works fine. So possible a regression introduced between these versions.

Strangely if I checkout both tags OC 8.0.3RC1 in core and calendar, the problem also appears.

PVince81:
After setting up v8.0.3 RC 1 (core + calendar) I saw that it works again.
But when doing so with v8.0.3 RC 2 it breaks.
I'll try and bisect it.

PVince81:
@georgehrke this is a regression introduced by "fix event sharing" backport: 7cadce0f675bf871eb641e70a99b1dc509bdd164

Please fix or revert. Failure to do so will cause the calendar app to break with the next 8.0.3 release.

PVince81:
I'll prepare a revert PR, you can then decide to merge or not.

PVince81:
Revert PR is here: https://github.com/owncloud/calendar/pull/770

georgehrke:
@PVince81 you just tried creating an event in your private calendar, not a shared one, right?

PVince81:
Yes, it was just plain old private calendar.

georgehrke:
ok, thx, looking into it

[...]

PR #772, opened 2015-04-16
fix event sharing on stable8

georgehrke:
fixes https://github.com/owncloud/calendar/issues/769

@PVince81 Would you mind testing this?


PVince81:
:+1: events work again.

Shared an event with a user, the user can see it. But the "class" dropdown is empty, maybe a separate issue ?

georgehrke:
> But the "class" dropdown is empty, maybe a separate issue ?

Not sure if I can follow.
Do you mean the "show full event", "show busy only", "hide event"?

georgehrke:
Never mind, was able to reproduce it.

PR #793, opened 2015-05-20
Fixing PHPUNIT execution - remove dependency to Contacts app

DeepDiver1975:
@georgehrke @raghunayyar please have a look - THX

NOTE:
- [ ] OCP\IAddressBook misses the method lastModified() - I'll take care and make up my mind on how to implement this

### DIFF
--- a/backend/contact/object.php
+++ b/backend/contact/object.php
@@ -120,6 +120,10 @@ class Object extends Contact implements BackendUtils\IObjectAPI {
 		$addressBooks = $this->contactsManager->getAddressBooks();
 		/** @var IAddressBook $addressBook */
 		foreach($addressBooks as $addressBook) {
+			if (!($addressBook instanceof IAddressBook)) {
+				continue;
+			}
+
 			$contacts = $addressBook->search('', ['FN', 'ANNIVERSARY', 'BDAY'], []);
 			foreach($contacts as $contact) {
 				if (!is_null($offset) && $i < $offset) {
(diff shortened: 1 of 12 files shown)


georgehrke:
Thx :blush: 
I'll check asap

backend/contact/object.php
```diff
@@ -203,19 +206,19 @@ private function getObjectFromContact(AddressBookEntity $addressBook, ContactEnt
...
+			'uid' => $contact['ID'],
 			'rrule' => 'FREQ=YEARLY'
 		]);
 
-		$object = $this->factory->createEntity($vobject, ObjectFactory::FORMAT_VObject);
+		$object = $this->factory->createEntity($vObject, ObjectFactory::FORMAT_VObject);
 		$object->setCalendar($this->calendar);
 		$object->setUri($this->getUriFromContact($addressBook, $contact));
-		$object->setEtag($contact->getETag());
+		$object->setEtag($contact['ETAG']);
```
georgehrke, line 225: is ETAG always set?

georgehrke:
looks good to me otherwise, thx for helping with rework :blush:

backend/contact/object.php
DeepDiver1975, line 225: that's something we need to double check

DeepDiver1975:
@georgehrke let's move this in - better then nothing :speak_no_evil:

georgehrke:
If i remember correctly this broke the backend api because some undefined methods were called, I'll report back asap.

PR #814, opened 2015-06-16
Fix calendar deletion while deleting users by command line

lmm-git:
This patch fixes the calender deletion while deleting users by command line or cron.

I did not found a proper method whether to detect if this function is called by the command line or cron job but I think this changes should solve the problem too because the userId is set always except of calls from command line or cron.


georgehrke:
code changes look good, will test asap

georgehrke:
@lmm-git thx for the patch :smile:

PR #818, opened 2015-06-23
[stable8] Add min and max version

LukasReschke:


georgehrke:
Is that required now?

georgehrke:
code looks good, will merge once tested

LukasReschke:
See https://github.com/owncloud/core/issues/17103 - this is there to hopefully prevent a lot of frustrated users once we released.

PR #820, opened 2015-06-23
Fix phpdoc of addCalendar()

Henni:


georgehrke:
:+1: 
thx for the fix :)

PR #867, opened 2015-07-19
Fix appname string

raimund-schluessler:
Fix the appname string under which a debug message regarding permissions is logged.


georgehrke:
thx :)

PR #870, opened 2015-07-21
[master] Clarify permission checks

LukasReschke:


gvde:
Wouldn't it be better to use getPermissions to get the permissions instead of "simulating" it there?

SergioBertolinSG:
:+1:

Issue #868, opened 2015-07-19
Lightning Sync: Only whole day events supported


Xarfay:
Hi,
I just installed Owncloud 8.1 with the up to date calendar app. In Owncloud 7 Calendar Sync with Lightning worked quite well. Right now it only syncs events that span the whole day. Is this a known bug? Is it possible that I configured something wrong? In addition sync with CalDay Sync in Android works perfectly.

Any help is appreciated!

Regards
Xarfay

georgehrke:
please use the [issue template](https://raw.githubusercontent.com/owncloud/core/master/issue_template.md) as described in our [contributing guidelines](https://github.com/owncloud/core/blob/master/CONTRIBUTING.md)

Xarfay:
### Steps to reproduce
1. Install Owncloud 8.1
2. Activate Calender Plugin from Appstore
3. Sync calendar within thunderbird using lightning plugin

### Expected behaviour
The calendar should sync all events and tasks

### Actual behaviour
- Tasks are synched
- Only (but all) whole-day events are shown in thunderbird

### Server configuration
Hosted server (all-inkl.com)

**Operating system**:
?Linux

**Web server:**
?Apache

**Database:**
MySQL 5.5.43

**PHP version:**
5.4.42

**ownCloud version:** (see ownCloud admin page)
8.1
**Updated from an older ownCloud or fresh install:**
fresh install
**List of activated apps:**
pre-installed + calendar, tasks, contacts, news and encryption
```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder
- No access available
```

**The content of config/config.php:**
<?php
$CONFIG = array (
  'instanceid' => 'XX',
  'passwordsalt' => 'XX',
  'secret' => 'XX',
  'trusted_domains' =>
  array (
    0 => 'XX.XX.com',
  ),
  'datadirectory' => '/www/htdocs/user/safe/data',
  'overwrite.cli.url' => 'http://XX.XX.com',
  'dbtype' => 'mysql',
  'version' => '8.1.0.8',
  'dbname' => 'XX',
  'dbhost' => 'XX',
  'dbtableprefix' => 'oc_',
  'dbuser' => 'XX',
  'dbpassword' => 'XX',
  'logtimezone' => 'UTC',
  'installed' => true,
  'mail_smtpmode' => 'php',
);
```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system --public
[...]

PR #876, opened 2015-07-22
Removed line causing wrong DTEND value

### LINKED ISSUE #849: ICS-Import sets DTEND to wrong value
When importing a ICS-File like the following:

    BEGIN:VCALENDAR
    VERSION:2.0
    PRODID:ownCloud Calendar 0.7.1
    X-WR-CALNAME:Import
    BEGIN:VEVENT
    UID:kt11261
    SUMMARY:My appointment
    DTSTART;TZID=Europe/Vienna:20150711T130000
    DTEND;TZID=Europe/Vienna:20150711T180000
    END:VEVENT
    END:VCALENDAR

the import produces:

    BEGIN:VCALENDAR
    VERSION:2.0
    PRODID:ownCloud Calendar 0.7.1
    BEGIN:VEVENT
    UID:kt11261
    SUMMARY:My appointment
    DTSTART;TZID=Europe/Vienna:20150711T130000
    DTEND;TZID=Europe/Vienna;VALUE=:20150711T180000
    END:VEVENT
    END:VCALENDAR

Please look at the DTEND-line! The Web-based ownCloud-calendar seems to be generous enough to display the event but when I access it trough another CalDAV client (in my case: Lightning) it is NOT displayed.

When editing the event in the Web-calendar the VCAL is rewritten as (have a look at the changed DTEND-setting!):

    BEGIN:VCALENDAR
    VERSION:2.0
    PRODID:ownCloud Calendar 0.7.1
    BEGIN:VEVENT
    UID:kt11261
    SUMMARY:Praxis BDA Intensiv
    DTSTART;TZID=Europe/Vienna:20150711T130000
    DTEND;TZID=Europe/Vienna:20150711T180000
    LAST-MODIFIED:20150712T003309Z
    DTSTAMP:20150712T003309Z
    CLASS:PUBLIC
    LOCATION:
    DESCRIPTION:
    CATEGORIES:
    END:VEVENT
    END:VCALENDAR

Which now can be read by Lightning etc.

irgendwie:
Fixes #849 and #868.
This line caused a new parameter to be created in `$objects->DTEND` with the name 'VALUE', which never gets used afterwards.
In the `serialize()` method this lead to the not needed "VALUE=" string in each event.

@msedv @Xarfay @polcape Does this fix your problems?

### DIFF
--- a/lib/import.php
+++ b/lib/import.php
@@ -126,7 +126,6 @@ class OC_Calendar_Import{
 			if(!is_null($object->DTSTART)){
 				$dtend = OC_Calendar_Object::getDTEndFromVEvent($object);
 				if($object->DTEND) {
-					$object->DTEND['VALUE'] = $object->DTSTART['VALUE'];
 					$dtend = $dtend->getDateTime();
 					$dtend->setTimeZone($object->DTSTART->getDateTime()->getTimeZone());
 					$object->DTEND->setDateTime($dtend);


irgendwie:
Also cc @PVince81, because this line was introduced by you (https://github.com/owncloud/calendar/blame/master/lib/import.php#L129)

lib/import.php
```diff
@@ -126,7 +126,6 @@ public function import() {
 			if(!is_null($object->DTSTART)){
 				$dtend = OC_Calendar_Object::getDTEndFromVEvent($object);
 				if($object->DTEND) {
```
georgehrke, line 128: off topic: shouldn't this be !$object->DTEND

irgendwie, line 128: @georgehrke What is the intention of this whole block?
https://github.com/owncloud/calendar/blob/master/lib/import.php#L128-L133
1. Check if there isn't a valid end date and add it as the same value as the start date?
2. Proper handling of date/time conversion stuff?
3. Something completely different?

polcape:
Now import works and the "VALUE=:" there isn't. But importing event with, for example, 

> DTSTART:20140716
> DTEND:20140717

it become with

> DTSTART;VALUE=DATE:20140716
> DTEND:20140717T000000Z

So, I think, it add the time on end date (and this is not correct) and it add "VALUE=" on star date but not on end date.

If I remove the all block, it works fine

if(!is_null($object->DTSTART)){
                $dtend = OC_Calendar_Object::getDTEndFromVEvent($object);
                if($object->DTEND) {
                    $dtend = $dtend->getDateTime();
                    $dtend->setTimeZone($object->DTSTART->getDateTime()->getTimeZone());
                    $object->DTEND->setDateTime($dtend);
                }
            }

lib/import.php
georgehrke, line 128: I don't think removing the line is a proper solution here. The if-block should be changed from `if($object->DTEND) {` to `if(!$object->DTEND) {`. Then this whole block makes sense. Do you agree?

gvde, line 128: I would say the block sets the time zone of DTEND to the same time zone as DTSTART, so that both are in the same time zone in the end.

Copying the VALUE parameter would make sure that both DTSTART and DTEND are of the same time, i.e. either both DATE or both DATETIME and not a mix of both. Of course, copying should only happen if the VALUE parameter is actually present in DTSTART.

Negating the condition to (!object->DTEND) would be useful, too, considering what getDTEndFromVEvent will do in that case. getDTEndFromVEvent returns a clone of DTSTART or even add DURATION if present. The block would then set a new DTEND into the event. But then I don't really get the other lines considering what getDTEndFromVEvent has already done...

Looking at an older commit 9c3aa7037c5f9c0f1d7a33cc4b88fa6a34de0cbc it seems as if the idea of the block from the beginning is to make sure the DTEND uses the same (VALUE) type and time zone as DTSTART. I do wonder, however, why it looks like it does at the moment...

irgendwie:
Closing in favor of #1004!

PR #884, opened 2015-08-04
Sort list of calendars by calendarorder

intrixius:
basic solution to fix request https://github.com/owncloud/calendar/issues/742
(No GUI implementation, so currently only possible to change the calendar order in the database table 'clndr_calendars')

### DIFF
--- a/lib/calendar.php
+++ b/lib/calendar.php
@@ -43,7 +43,7 @@ class OC_Calendar_Calendar{
 			$active_where = ' AND `active` = ?';
 			$values[] = (int)$active;
 		}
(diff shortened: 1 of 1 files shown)


intrixius:
at first this seemed to work, but now I'm adding more calendars (with a corresponding `calendarorder`), but somehow they are not sorted anymore correctly...
Is there any other place where it gets resorted?

Edit: it seems that the sorting only happens on owned calendars, not shared ones...

intrixius:
I have changed the way of sorting, now also shared calendars are sorted

DeepDiver1975:
From my pov sorting the calendars is not a server side job but something to be done on the client side.

Issue #832, opened 2015-07-07
[oC 8.1] can no longer sync shared calendars


schiesbn:
I just updated my server to the 8.1 branch, both master and calendar. The update went well but my mobile phone can no longer sync a calender which was shared with me, the calendar is shared with all permissions. I have this errror in the owncloud.log:

````
[...]

PR #885, opened 2015-08-06
Overwrite getMultipleChildren

gvde:
Overwrite getMultipleChildren to create calendar objects of OC_Connector_Sabre_CalDAV_CalendarObject which return the correct ACL for shared calendars.

### DIFF
--- a/lib/sabre/calendar.php
+++ b/lib/sabre/calendar.php
@@ -129,4 +129,21 @@ class OC_Connector_Sabre_CalDAV_Calendar extends \Sabre\CalDAV\Calendar {
 
 	}
 
+	/**
+	 * This method receives a list of paths in it's first argument.
+	 * It must return an array with Node objects.
+	 *
+	 * If any children are not found, you do not have to return them.
+	 * 
+	 * @return array
+	 */
+	public function getMultipleChildren(array $paths) {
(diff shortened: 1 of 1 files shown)


georgehrke:
Thanks a lot for the fix, will review asap!

georgehrke:
fixes https://github.com/owncloud/calendar/issues/832

lib/sabre/calendar.php
```diff
@@ -129,4 +129,21 @@ public function getChildren() {
...
+	 * This method receives a list of paths in it's first argument.
+	 * It must return an array with Node objects.
+	 *
+	 * If any children are not found, you do not have to return them.
+	 * 
+	 * @return array
+	 */
+	public function getMultipleChildren(array $paths) {
+		$objs = $this->caldavBackend->getMultipleCalendarObjects($this->calendarInfo['id'], $paths);
+		$children = [];
```
georgehrke, line 142: @DeepDiver1975 We dropped PHP 5.3 support with 8.1, right?

georgehrke:
patch itself works :+1:

lib/sabre/calendar.php
georgehrke, line 142: Never mind, admin man says `PHP 5.4 + required`

PR #919, opened 2015-09-01
Fix some tests, typos

codeling:
fix slugify test: on my system as well as on travis, "äöü" gets converted to "aou", not "aeoeue" (except if I manually change the locale to de_AT.utf8 first).


georgehrke:
:+1:
Thx :)

PR #920, opened 2015-09-01
Collection test

codeling:
Introduces unit tests for collection class.
In the course of writing the tests I came across peculiar behaviour in the addCollection / addObjects methods, which I think is a bug; the tests check what the docs say, and therefore fail for the current version. Also, function noDuplicates cannot be called (reports "cannot convert IEntity to string), therefore I suspect this method not to be used anywhere.

I have, in a separate commit, added fixes & simplifications for the addCollection/addObjects class, please check if these make sense.

### DIFF
--- a/db/collection.php
+++ b/db/collection.php
@@ -84,20 +84,11 @@ abstract class Collection implements ICollection {
 	 *
 	 * @param IEntity $object entity to be added
 	 * @param integer $nth insert at index, if not set, entity will be appended
+	 *        if negative, it will be inserted at the position that many elements from the end
 	 * @return $this
 	 */
 	public function add(IEntity $object, $nth=null) {
-		if ($nth === null) {
-			$this->objects[] = $object;
-		} else {
-			$objects = array_slice($this->objects, 0, $nth, true);
-			$objects = array_merge($objects, array($object));
-			$objects = array_merge($objects, array_slice($this->objects, $nth, $this->count(), true));
-
-			$this->objects = $objects;
-		}
-
-		return $this;
+		return $this->addObjects(array($object), $nth);
 	}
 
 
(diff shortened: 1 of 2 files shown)


db/collection.php
```diff
@@ -301,7 +282,9 @@ public function offsetGet($offset){
 	 * @param mixed $value
 	 */
 	public function offsetSet($offset, $value){
+		// TODO: why not IEntity in parameter list?
 		if ($value instanceof IEntity) {
+			// TODO: why this - shouldn't setObjects be used in that case?
```
georgehrke, line 285: this should indeed use setObjects

db/collection.php
```diff
@@ -79,25 +79,17 @@ public static function fromArray(array $entities) {
 	}
 
 
+	// instead of duplication!
```
georgehrke: please remove this comment,
cleaning up `add`, `addCollection` and `addObjects` makes a lot of sense though :)

db/collection.php
```diff
@@ -301,7 +282,9 @@ public function offsetGet($offset){
 	 * @param mixed $value
 	 */
 	public function offsetSet($offset, $value){
+		// TODO: why not IEntity in parameter list?
```
georgehrke: PHP Fatal error:  Declaration of OCA\Calendar\Db\Collection::offsetSet() must be compatible with ArrayAccess::offsetSet($offset, $value) in /var/www/apps/calendar/db/collection.php on line 27

:wink:

tests/unit/db/collectionTest.php
```diff
@@ -21,7 +21,350 @@
...
+
+		$coll = ConcreteCollection::fromEntity($mockedValid);
+		$this->assertTrue($coll->isValid());
+
+		$coll->add($mockedInvalid);
+		$this->assertFalse($coll->isValid());
+	}
+
+	/*
+	public function testNoDuplicates() {
```
georgehrke, line 309: Why is this method commented out?

georgehrke:
replaced by https://github.com/owncloud/calendar-rework/pull/25

PR #921, opened 2015-09-01
Unit Tests: Db\Entity

codeling:
Those tests were rather simple and no bugs detected ;).
But I'm not sure if the multiple small classes deriving from Entity are a good idea or what could be a better solution there - I don't think they can be mocked, as registerMandatory and registerTypes are called during construction already and need to be overriden...


georgehrke:
:+1:
thx a lot :)

PR #938, opened 2015-09-22
Use rawurlencode to encode spaces in user names

gvde:
Use rawurlencode to encode spaces in user names as %20 instead of + to be put into the iOS CalDAV url in the frontend.
The + encoding is ambiguous and it will be considered part of the user name if passed to caldav. Thus it looks for user 'First+Last' instead of 'First Last'.

### DIFF
--- a/templates/calendar.php
+++ b/templates/calendar.php
@@ -106,7 +106,7 @@
 				</li>
 				<li>
 					<label class="bold"><?php p($l->t('iOS/OS X CalDAV address')); ?></label>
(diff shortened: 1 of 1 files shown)


ghost:
Another approach is https://github.com/owncloud/calendar/pull/945 using the same syntax as already used within the calendars "CalDAV link": https://github.com/owncloud/calendar/blob/master/templates/part.choosecalendar.rowfields.php#L28

No idea which one is better or which should be used.

gvde:
@RealRancor I didn't see that. Using the same way is probably better even if I don't quite understand it. I wonder why quotes (and only quotes) should be entity encoded in the user name. I hope someone thought about it when this was done or otherwise it would need to be revisited...
I am closing this...

PR #941, opened 2015-09-23
Fix checkboxes

Henni:
Reorder labels of checkboxes to work with the new checkbox styles (#18735).

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -199,6 +199,11 @@ button.category{margin:0 3px;}
 	display: block;
 	margin: 5px 0;
 }
+#event-time:after {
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
Is this the way we want this to looks like?

bildschirmfoto von 2015-10-12 12-36-24

Henni:
@DeepDiver1975 Probably not. I'll take another look.

Henni:
image

now looks like this and should fallback to the normal checkboxes in 8.1 and before.

DeepDiver1975:
THX :+1:

PR #944, opened 2015-09-28
Added setting to enable/disable contact birthdays

ftiede:
By default calendar app exports a special calendar containing all
contact's birthdays. For users who use clients with their own linking
between addressbooks and calendars, f.ex. Thunderbird, kontact, Android
with Birthday Calendar adapter, this produces overhead.

This change introduces a global setting inside ownCloud's calendar app
to disable this feature.

### DIFF
--- a/ajax/settings/contactbirthdays.php
+++ b/ajax/settings/contactbirthdays.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * Copyright (c) 2011, 2012 Georg Ehrke <ownclouddev at georgswebsite dot de>
+ * Copyright (c) 2015 Felix Tiede <info at pc-tiede dot de>
+ * This file is licensed under the Affero General Public License version 3 or
+ * later.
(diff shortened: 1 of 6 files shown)


DeepDiver1975:
From my POV in such a scenario the user simple should not add the birthday calendar to that specific client.

ftiede:
Unfortunately there are clients where that is not precisely possible. And I can think of scenarios - more than just ownCloud as a contact source - where the birthday calendar is required and addressbooks can't be chosen.

From my POV it was/is easier to add that setting to oc/calendar than to hack all the client/adapters out there.

DeepDiver1975:
Point taken - but then this is a client specific setting - not a user specific - right?

ftiede:
Depends. My setup alone runs three different client/adapter combinations I know of.

So yeah, it would be client specific, yet as I don't see how to chain this setting to the specific client used for synchronization I opted for the user to select what he likes best for his combination(s).

Issue #937, opened 2015-09-21
Problem with Mac OS X Calendar and space between username


reinerj:
We face a problem with OC 8.1.2 and MAC OS X Calendar It works perfect fine with Evolution Calendar on Linux but under Mac we get the following problem. 

The username on OC is something like „Hugo Boss“ with a space. between the names. The authentication fail. When we create a new test user without the space the login works perfect. We cannot simple change the usernames as OC is in production since around 3 years. 

Mac OS X is 10.10.5

georgehrke:
please use the [issue template](https://raw.githubusercontent.com/owncloud/core/master/issue_template.md) as described in our [contributing guidelines](https://github.com/owncloud/core/blob/master/CONTRIBUTING.md) ......

reinerj:
### Steps to reproduce
1. Create a user with s username with a space between first and last name. Something like „Hugo Boss"
2. Add a new Caldav account and copy the URL from OC 
3. After you save you get an error message that <calendar cannot connect to OC and you asked for the password again and again 

### Expected behaviour
Calendar should be added, Works since a long time on Evolution + Fedora

### Actual behaviour
Nothing beside you asked again and again for the password. The URL which is shown is like: https://mydomain.org/caldav/principals/hugo+boss

### Server configuration
**Operating system**:
OpenBSD, We tried the same with CentOS

**Web server:**
Apache
**Database:**

MySQL
**PHP version:**
5.5

**ownCloud version:** (see ownCloud admin page)
8.1.2

**Updated from an older ownCloud or fresh install:**

**List of activated apps:**
Activity
Default encryption module
Deleted files
Pictures 
Share Files
Text Editor
Versions
Calendar


```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder
```

**The content of config/config.php:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system --public
from within your ownCloud installation folder

or 

Insert your config.php content here
(Without the database password, passwordsalt and secret)
```

**Are you using encryption:** yes

**Are you using an external user-backend, NO

### Client configuration
Mac OS X Calendar

**Operating system:**
Mac OS X 10.10.5

#### ownCloud log (data/owncloud.log)
```
[...]

PR #945, opened 2015-09-29
Fix URL encoding for iOS/OS X CalDAV address

ghost:
Proposal for problem described in https://github.com/owncloud/calendar/issues/937

### DIFF
--- a/templates/calendar.php
+++ b/templates/calendar.php
@@ -106,7 +106,7 @@
 				</li>
 				<li>
 					<label class="bold"><?php p($l->t('iOS/OS X CalDAV address')); ?></label>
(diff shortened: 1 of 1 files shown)


gvde:
This is probably better than my initial #938.

ghost:
Copied over from: https://github.com/owncloud/calendar/pull/938#issuecomment-143981507

> @RealRancor I didn't see that. Using the same way is probably better even if I don't quite understand it. I wonder why quotes (and only quotes) should be entity encoded in the user name. I hope someone thought about it when this was done or otherwise it would need to be revisited...

DeepDiver1975:
@LukasReschke encoding-fu - THX

DeepDiver1975:
tested with cadaver :+1:

PR #952, opened 2015-10-12
Public link sharing

DeepDiver1975:
@jancborchardt let's test this and get this finally in for 8.2

### DIFF
--- a/ajax/event/edit.form.php
+++ b/ajax/event/edit.form.php
@@ -15,7 +15,7 @@ $link_shared = false;
 if(!OCP\User::isLoggedIn()) {
 
 	// do we have a token?
-	if (!\OC::$session->exists('public_link_token')) {
+	if (!\OC::$server->getSession()->exists('public_link_token')) {
 		// nope, bail out!
 		OCP\User::checkLoggedIn();
 	}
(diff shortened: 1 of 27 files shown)


jancborchardt:
Oh nice! Please also review @rysiekpl @georgehrke @raghunayyar @schiesbn! :)

schiessle:
How should this work exactly? I created a link and opened in the browser and then I can see the complete calendar. I'm also able to import it into my local calendar?

Some issues I discovered during a short test:

Public Link View:
- [ ] if no user (or a different user) is logged in the calendar is empty. I only see the events as long as the owner of the calendar is logged in while I visit the public link
- [ ] the shared calendar still show all the control elements (share, download, link, rename, delete) some of them simply doesn't work, some redirect me to the login page and some show a inline error message

Owner View:
- [ ] the share dialog seems to be broken. No drop-down (like on master) but permanently visible (also elements overlap if you create new calendars, etc.)
- [ ] after page reload the information about the link share is gone

MorrisJobke:
- [ ] Sidebar looks broken:

bildschirmfoto von 2015-10-13 11-33-33

DeepDiver1975:
> How should this work exactly? I created a link and opened in the browser and then I can see the complete calendar.

Ah - full calendars can be shared as well - I was only testing sharing of a single event

DeepDiver1975:
@MorrisJobke I guess the sidebar is broken because of we did move the share dialog into the app ....

DeepDiver1975:
So .... what shall we do? I honestly don't have more time to invest into this.

Declare this change as total failure? Quite common in this app .... :imp:

schiessle:
> Declare this change as total failure?

probably that's the only feasible decision.

Generally speaking, in Berlin @georgehrke and @raghunayyar accepted the challenge to get the calendar re-write ready for ownCloud 9.0. If this plan work out, all work we put now in features for the old calendar will be lost is a few weeks/months. So it is probably not worth the efforts.

DeepDiver1975:
> Generally speaking, in Berlin @georgehrke and @raghunayyar accepted the challenge to get the calendar re-write ready for ownCloud 9.0.

All beers are on me if this really happens ...

jancborchardt:
> > Generally speaking, in Berlin @georgehrke and @raghunayyar accepted the challenge to get the calendar re-write ready for ownCloud 9.0.
> 
> All beers are on me if this really happens ...

Saving for later. Now you better manage this @georgehrke @raghunayyar or we’ll all go thirsty! :D

PR #354, opened 2014-02-28
Update backend.php

ednt:
Fixes #350

Not possible to create a calendar via http.

PR #953, opened 2015-10-13
Fixes #350 - Not possible to create a calendar via http.

### LINKED ISSUE #350: Not possible to create a calender via http
Hi,

now I'm totally surprised:
I already fixed the contact stuff (together with thangus) and now it's the same for calendar.

It is not possible to create a calendar from 'outside' via html.

Since I know now where I have to look, I found out very fast, that the function

createCalendar() in lib/sabre/backend.php

is a total 'disaster'.
All parameters are put into $values and later $newValues is used, which is never defined before
Next point:
All keys were done with a ':' in front, but later the access is without ':'

To make it short:

Here is a working function:

        public function createCalendar($principalUri,$calendarUri, array $properties) {
                $values = array(
                        'principaluri' => $principalUri,
                        'uri'             => $calendarUri,
                        'ctag'           => 1,
                );

                // Default value
                $sccs = '{urn:ietf:params:xml:ns:caldav}supported-calendar-component-set';
                if (!isset($properties[$sccs])) {
                        $values['components'] = 'VEVENT,VTODO';
                } else {
                        if (!($properties[$sccs] instanceof Sabre_CalDAV_Property_SupportedCalendarComponentSet)) {
                                throw new Sabre_DAV_Exception('The ' . $sccs . ' property must be of type: Sabre_CalDAV_Property_SupportedCalendarComponentSet');
                        }
                        $values['components'] = implode(',',$properties[$sccs]->getValue());
                }

                foreach($this->propertyMap as $xmlName=>$dbName) {
                        if (isset($properties[$xmlName])) {
                                $values[$dbName] = $properties[$xmlName];
                        }
                }

                if(!isset($values['displayname'])) $values['displayname'] = 'unnamed';
                if(!isset($values['components'])) $values['components'] = 'VEVENT,VTODO';
                if(!isset($values['timezone'])) $values['timezone'] = null;
                if(!isset($values['calendarorder'])) $values['calendarorder'] = 0;
                if(!isset($values['calendarcolor'])) $values['calendarcolor'] = null;
                if(!is_null($values['calendarcolor']) && strlen($values['calendarcolor']) == 9) {
                        $values['calendarcolor'] = substr($values['calendarcolor'], 0, 7);
                }

                return OC_Calendar_Calendar::addCalendarFromDAVData($principalUri,$calendarUri,$values['displayname'],$values['components'],$values['timezone'],$values['calendarorder'],$values['calendarcolor']);
        }

And I think $fieldname can totally removed and  $myValue too
They makes no sense, but I'm not familar with php, please check this.
I checked it with my code: it works without them, so I removed the lines.

Best regards

Bernd

And greetings to thangus!

DeepDiver1975:
rebased version of https://github.com/owncloud/calendar/pull/354

### DIFF
--- a/lib/sabre/backend.php
+++ b/lib/sabre/backend.php
@@ -101,48 +101,54 @@ class Backend extends \Sabre\CalDAV\Backend\AbstractBackend {
 	 * @throws \Sabre\DAV\Exception
 	 */
 	public function createCalendar($principalUri,$calendarUri, array $properties) {
-		$fieldNames = array(
-			'principaluri',
-			'uri',
-			'ctag',
-		);
-		$values = array(
-			':principaluri' => $principalUri,
-			':uri'		  => $calendarUri,
-			':ctag'		 => 1,
-		);
+		
+		$values = array();
 
 		// Default value
 		$sccs = '{urn:ietf:params:xml:ns:caldav}supported-calendar-component-set';
-		$fieldNames[] = 'components';
 		if (!isset($properties[$sccs])) {
-			$values[':components'] = 'VEVENT,VTODO';
+			$values['components'] = 'VEVENT,VTODO';
 		} else {
 			if (!($properties[$sccs] instanceof \Sabre\CalDAV\Property\SupportedCalendarComponentSet)) {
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
tested :+1:

PR #976, opened 2015-11-03
Prevent hour 24 and minute 60

phil-davis:
It seems to me that checkTime will currently allow a time like "09:60" or "24:00" or "24:01" or even "24:60"
Maybe 24:00 is allowable? Does that indicate the end of a day, as compared to the same time "00:00" of the next day?

But minutes of "60" would always be invalid. And a "24" hour with minutes other than "00" is also invalid.


phil-davis:
This ancient and (I guess) is not the current code base - closing.

PR #978, opened 2015-11-04
Consolidation of Colors in Calendar Color Picker

EtlamGit:
Used a standard Color Wheel with 8 segments + Gray to generate base colors.
Variants are tinted (more pastel) from the base color. Tinting is similar to the last commit.
Reordered colors to better match the 9 wide layout:
colorpicker


EtlamGit:
I would prefer https://github.com/owncloud/calendar/pull/979

So feel free to reject this one or to leave comments.

PR #979, opened 2015-11-05
12 Segment Color Wheel for Calendar Color Picker

EtlamGit:
This is an alternative to https://github.com/owncloud/calendar/pull/978 Obviously only one of them should be merged.
This time used a standard Color Wheel with 12 segments + Gray to generate base colors. Variants are tinted (more pastel) and darkened from the base color. Darkening and Tinting are possible in any shade, so more or less variants are possible easily.
Changed to 13 wide layout to get all colors from the wheel in one row:
colorpicker12


EtlamGit:
All defined colors:
colorpicker12_10

Default colors resulting from commenting out some of them:
colorpicker12_20b

PR #994, opened 2015-11-27
only set DTEND VALUE if DTSTART has VALUE set.

gvde:
Set VALUE parameter of DTEND to the same as DTSTART but only if DTSTART has the parameter set.


georgehrke:
code looks good, need some testers here.

gvde:
new backport to stable8.2 in #1004
do you need backport to stable8.1, too?

cecilebraun:
A backport for stable 8.1 will be wonderful.

gvde:
Backport to 8.1 in #1022

PR #1004, opened 2015-12-11
only set DTEND VALUE if DTSTART has VALUE set

gvde:
Set VALUE parameter of DTEND to the same as DTSTART but only if DTSTART has the parameter set.
Backport of #994 to stable8.2


tlhackque:
This solves #1023, which is a serious synchronization issue.

It would be good if it could be merged.

PR #1010, opened 2015-12-31
Fixes #1009: add LAST-MODIFIED to calendardata when it is absent

### LINKED ISSUE #1009: ownCloud 8.2.2.2 and calendar 0.8.1: Call to a member function setDateTime() on null in /var/www/owncloud/apps/calendar/ajax/event/move.php on line 38
Dear devs,

I am getting the following error each time I am trying to move some event from one day to another in the Calendar:

```
2015/12/31 09:27:52 [error] 8#8: *703 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Call to a member function setDateTime() on null in /var/www/owncloud/apps/calendar/ajax/event/move.php on line 38" while reading response header from upstream, client: [--REMOVED--], server: myowncloud.com, request: "POST /index.php/apps/calendar/ajax/event/move.php HTTP/2.0", upstream: "fastcgi://unix:/var/run/shared/php5-fpm-myowncloud.com.sock:", host: "myowncloud.com"
```

```
server:/$ cat /var/www/owncloud/apps/calendar/appinfo/version
0.8.1
```

Please advise what could this be? Any workaround to this?

ghost:
to close https://github.com/owncloud/calendar/issues/1009

Suggestions/critics are welcomed as my php experience is very little :-)

### DIFF
--- a/ajax/event/move.php
+++ b/ajax/event/move.php
@@ -35,7 +35,7 @@ unset($vevent->DURATION);
 
 $now = new DateTime('now');
 $now->setTimeZone(new \DateTimeZone('UTC'));
-$vevent->__get('LAST-MODIFIED')->setDateTime($now);
+$vevent->__set('LAST-MODIFIED', clone $now);
 $vevent->DTSTAMP = $now;
 
 try {


ajax/event/move.php
```diff
@@ -35,6 +35,10 @@
 
 $now = new DateTime('now');
 $now->setTimeZone(new \DateTimeZone('UTC'));
+$lastModified = $vevent->__get('LAST-MODIFIED');
+if (is_null($lastModified)) {
+	$lastModified = $vevent->add('LAST-MODIFIED');
+}
 $vevent->__get('LAST-MODIFIED')->setDateTime($now);
```
gvde: Shouldn't a simple change of 

```
$vevent->__get('LAST-MODIFIED')->setDateTime($now);
```

to

```
$vevent->__set('LAST-MODIFIED', $now);
```

be enough? That should be the identical rewrite of what happens with DTSTAMP in the line after...

gvde:
The problem with the LAST-MODIFIED may be present at other places as well. Should be checked everywhere if it's O.K. at all places...

ajax/event/move.php
ghost: Thx @gvde ! I've tested and it is working for me.

gvde:
There may be one problem assigning the same object to two properties: if the object is changed through one property it affects the other property as well (e.g. if the following code would somewhere modify the time zone of DTSTAMP to something else it would also change the timezone of LAST-MODIFIED).

Thus, I guess to be safe, you should assign LAST-MODIFIED not $now but "clone $now" which creates a copy of the object in $now. Then DTSTAMP and LAST-MODIFIED both are set to the current time but using two different objects.

ghost:
@gvde thanks for your valuable comment! I've just pushed the update, please see if everything is okay now from your point of view.

gvde:
Looking good to me...

PR #1022, opened 2016-01-29
Set VALUE parameter of DTEND to the same as DTSTART but only if DTSTA…

gvde:
…RT has the parameter set.

Backport of #994 to stable 8.1


cecilebraun:
I tested the Code and it works.
oc.8.1
tb.38.5.0
evolution 3.8.5

